PR #2, opened 2020-02-03
feat: Staging mode switch

yehjxraymond:
Allow the app to switch into staging network once the icon is long pressed for 5 seconds.

### DIFF
--- a/src/components/Layout/AppName.tsx
+++ b/src/components/Layout/AppName.tsx
@@ -1,15 +1,31 @@
 import React, { FunctionComponent } from "react";
 import { color, fontSize } from "../../common/styles";
 import { AppText } from "./AppText";
+import { AppMode } from "../../common/hooks/useConfig";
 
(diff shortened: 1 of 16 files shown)


src/components/Layout/AppName.tsx
```diff
@@ -1,15 +1,31 @@
...
+        color: color("grey", 0),
+        fontFamily: "inter-bold",
+        fontSize: fontSize(4)
+      }}
+    >
+      Rationally
+    </AppText>
+    {mode === AppMode.staging ? (
+      <AppText
+        style={{ color: "red", fontWeight: "bold", fontSize: fontSize(3) }}
```
sebastianquek: `fontWeight: bold` doesn't work when using fonts, need to use `fontFamily: "inter-bold"` instead

yehjxraymond: Didnt know that, thanks!

PR #8, opened 2020-02-17
fix: use context for config

sebastianquek:
The existing `useConfig` returns a stale config while it waits for the async get call to resolve each time `useConfig` is 'mounted'. By relying on context, we can ensure that the config obtained from AsyncStorage is resolved before any of the context consumers mount/render. Subsequent calls to get/set the config relies on the context and the updating of AsyncStorage becomes a side effect.


src/context/config.tsx
```diff
@@ -0,0 +1,66 @@
...
+      [key]: value
+    };
+    setConfig(nextConfig);
+    AsyncStorage.setItem(CONFIG_KEY, JSON.stringify(nextConfig));
+  };
+
+  const loadConfigFromStore = async (): Promise<void> => {
+    const configStr = await AsyncStorage.getItem(CONFIG_KEY);
+    if (configStr) {
+      setConfig(JSON.parse(configStr));
```
Nebulis: there is one problem with this, if you add a new settings and there already exist a stored config, then there will be no default value provided for the new setting

=> https://github.com/Open-Attestation/identity-wallet/pull/63/files#diff-5689e8dfaa1efe9470061c2814ecf3e0R37

src/context/config.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+const CONFIG_KEY = "CONFIG";
+const DEFAULT_CONFIG: Config = { appMode: AppMode.production };
+
+const ConfigContext = createContext<ConfigContext>({
+  config: DEFAULT_CONFIG,
+  setConfigValue: () => {} // eslint-disable-line @typescript-eslint/no-empty-function
+});
+
+export const useConfigContext = (): ConfigContext =>
```
Nebulis: you could have kept useConfig as a name

src/context/config.tsx
sebastianquek: Ok will add this in

src/context/config.tsx
sebastianquek: I left the `Context` in there to remind the consumer not use this without ensuring an enclosed context.

PR #12, opened 2020-03-30
Add sorting logic and edit styling

ilhaoni:
Hello! 

What's in this PR:
- Added custom sorting logic (the order will be given from the backend in the `Name` key
- Styled the items again as the display name is now quite long

Do let me know if it is ok (:

### DIFF
--- a/src/components/CustomerQuota/CustomerQuotaScreen.tsx
+++ b/src/components/CustomerQuota/CustomerQuotaScreen.tsx
@@ -62,7 +62,7 @@ const styles = StyleSheet.create({
     flexDirection: "row",
     borderRadius: borderRadius(3),
     alignItems: "center",
-    minHeight: size(7),
+    minHeight: size(9),
     borderStyle: "dashed",
     borderWidth: 1,
     borderColor: color("grey", 20)
@@ -177,10 +177,28 @@ const CanBuyResult: FunctionComponent<{
       <CustomerCard nric={nric}>
         <View style={styles.resultWrapper}>
           {Object.entries(cart)
-            .sort()
+            .sort((itemOne, itemTwo) => {
+              const itemOneName = getProduct(itemOne[0])?.name || "";
+              const itemOneOrder = itemOneName.slice(
+                0,
+                itemOneName.indexOf(" ")
+              );
+              const itemTwoName = getProduct(itemTwo[0])?.name || "";
+              const itemTwoOrder = itemTwoName.slice(
+                0,
+                itemTwoName.indexOf(" ")
+              );
+
+              return parseInt(itemOneOrder) - parseInt(itemTwoOrder);
+            })
             .map(([category, canBuy]) => {
               const product = getProduct(category);
-              const categoryText = product?.name || category;
+              let categoryText;
+              if (product?.name) {
+                categoryText = product?.name.slice(product?.name.indexOf(" "));
+              } else {
+                categoryText = category.split("-");
+              }
               return canBuy === null ? (
                 <View style={styles.checkboxesListItem} key={category}>
                   <NoQuotaCategoryItem
--- a/src/components/CustomerQuota/CustomerQuotaScreen.tsx
+++ b/src/components/CustomerQuota/CustomerQuotaScreen.tsx
@@ -178,10 +178,10 @@ const CanBuyResult: FunctionComponent<{
         <View style={styles.resultWrapper}>
           {Object.entries(cart)
             .sort((itemOne, itemTwo) => {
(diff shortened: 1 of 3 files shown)


src/components/CustomerQuota/CustomerQuotaScreen.tsx
```diff
@@ -177,10 +177,28 @@ const CanBuyResult: FunctionComponent<{
       <CustomerCard nric={nric}>
         <View style={styles.resultWrapper}>
           {Object.entries(cart)
-            .sort()
+            .sort((itemOne, itemTwo) => {
```
ilhaoni: Added sorting logic here. Previously we were sorting by ascending ascii character. 

Currently, the backend will return the `name` field that will be prepended with a number (like `1. XXX`). This function will extract out the number and order it accordingly

src/components/CustomerQuota/CustomerQuotaScreen.tsx
```diff
@@ -177,10 +177,28 @@ const CanBuyResult: FunctionComponent<{
...
+                0,
+                itemTwoName.indexOf(" ")
+              );
+
+              return parseInt(itemOneOrder) - parseInt(itemTwoOrder);
+            })
             .map(([category, canBuy]) => {
               const product = getProduct(category);
-              const categoryText = product?.name || category;
+              let categoryText;
```
ilhaoni: As we prepend the number, we'll have to remove it before displaying it

src/components/CustomerQuota/CustomerQuotaScreen.tsx
sebastianquek: I'm wondering if we want to change the `Product` interface instead, and add the order there. This will make it easier to sort. We're likely to need it when sorting the purchased products too.

src/components/CustomerQuota/CustomerQuotaScreen.tsx
sebastianquek: As mentioned above, we can process whatever is given by the backend so that these components don't need to deal with it

src/components/CustomerQuota/CustomerQuotaScreen.tsx
ilhaoni: Ohhh, I see I see! Okk will make the changes!

src/components/CustomerQuota/CustomerQuotaScreen.tsx
ilhaoni: ok!

ilhaoni:
Currently the items will all be unselected by default also, thanks Seb for helping out :D

src/components/CustomerQuota/CustomerQuotaScreen.tsx
ilhaoni: Made the changes!

src/components/CustomerQuota/CustomerQuotaScreen.tsx
ilhaoni: Made the changes!

src/components/CustomerQuota/CustomerQuotaScreen.tsx
```diff
@@ -178,27 +178,14 @@ const CanBuyResult: FunctionComponent<{
...
-              const itemOneOrder = itemOneName.slice(
-                0,
-                itemOneName.indexOf(" ")
-              );
-              const itemTwoName = getProduct(itemTwo[0])?.name || "";
-              const itemTwoOrder = itemTwoName.slice(
-                0,
-                itemTwoName.indexOf(" ")
-              );
+              const ProductOneOrder = getProduct(itemOne[0])?.order || 0;
```
sebastianquek: May want to use camelCase here (:

src/components/CustomerQuota/CustomerQuotaScreen.tsx
sebastianquek: 👍

src/components/CustomerQuota/CustomerQuotaScreen.tsx
sebastianquek: 👍

src/components/CustomerQuota/CustomerQuotaScreen.tsx
ilhaoni: Ohh, ooops, thanks for the check. Will edit!

ilhaoni: Done :D

PR #13, opened 2020-03-31
Fix issue where user cannot select "Next Customer" at the purchase screen

ilhaoni:
Found an edge case while doing testing. Let's say there's an example where a customer comes in and buys everything, at the Purchase screen, the "Next Customer" button will not be within the screen as the purchase list is too long. Hence the adding of the paddingBottom to all ctaButtons instead, rather than just for one.

Other stuff in this PR:
- There was some error reported by Policy, so I fixed it
- Adding a console log for the authenticated for the time being to monitor the bug ZY raised


ilhaoni:
Ready for review~

PR #14, opened 2020-04-01
Update Expo to 36.0.0

ilhaoni:
Hello!

This is a PR to see if we can fix the bug found. (Can see in the chat). I mainly followed this guide here: https://blog.expo.io/expo-sdk-36-is-now-available-b91897b437fe

I also added a react-dev-tools as well so that it's easier for us to debug.

A few questions I have:
- Are we using expoKit? I didnt update this. 
- There's this instruction here which I didn't run it haha. Shouldn't need to do this right
```Update the Expo app on your phones from the App Store / Google Play. expo-cli will automatically update your apps in simulators if you delete the existing apps, or you can run expo client:install:ios and expo client:install:android.```

Do let me know if it's ok~


sebastianquek:
Let's close this for now, and upgrade to sdk37 in future (:

PR #19, opened 2020-04-05
fix: improve disabling of camera when screen not focused

sebastianquek:
There were reports of people who have not redeemed anything seeing that they don't have any quota left. I noticed this firsthand, and it occurs when a batch (household) of IDs were being scanned.

I managed to reproduce it several times when combined with a throttled network and on slower devices. For example, given 2 IDs, A and B, by scanning in quick succession, there are times when the quota of B is recorded when the user thinks A's quota is being recorded. So when scanning B, it correctly shows that B does not have quota left. If the user scans A again, he'll notice that there's quota available for A.

The issue stems from the camera on the scanning ID screen not being turned off properly after an ID was identified. So just before navigating to the quota available screen, the camera could pick up on the next ID (that has been placed in view of the camera). This causes the 2nd ID's quota to be shown. Since everyone's quota is the same, it's difficult to discern the differences in ID.

I've used React navigation's `withNavigationFocus` rather than relying on the event listeners. I've tested several times and it effectively stops the camera from scanning.

### DIFF
--- a/src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
+++ b/src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
@@ -8,7 +8,6 @@ import {
   ActivityIndicator
 } from "react-native";
 import { Feather } from "@expo/vector-icons";
-import { NavigationProps } from "../../types";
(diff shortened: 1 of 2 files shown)


ilhaoni:
Looks great haha :D

PR #20, opened 2020-04-05
feat: last transaction time

sebastianquek:
* If last transaction was more than 10mins ago, show full date
* If not, show duration (e.g. 5 minutes ago)


ilhaoni:
Awesome~~

PR #21, opened 2020-04-05
feat: style updates

sebastianquek:
* Ensure shadow of the customer card can be seen
* Remove input field from login screen, users must scan a QR.


ilhaoni:
Looks good~ ^^

PR #22, opened 2020-04-06
feat: add update available banner

sebastianquek:
Show banner when update is available and has been downloaded. Pressing update reloads the app using the latest downloaded version.

- When app is freshly opened / became in focus
    - If last sync time in AsyncStorage is more than 60 mins ago, check for updates, subsequently update AsyncStorage
- When app is kept open
    - Whenever the CollectCustomerDetailsScreen is focused, do the same as above

|Banner on login screen | Banner on the scanning screen|
|--|--|
|Simulator Screen Shot - iPhone 11 Pro - 2020-04-20 at 23 14 49|Simulator Screen Shot - iPhone 11 Pro - 2020-04-20 at 23 04 06|

---

### Previous remarks

~I see 2 problems,~
~1. When updating, users close and open the app too quickly. So the new bundle hasn't been fully downloaded.~
~2. Difficult to push updates without informing users manually.~

~For 1, we leave the fallbackToCacheTimeout as 0, and we listen to the started/finished downloading events to show progress and prompt the user to reload. In the longer term we can just increase it to 20 seconds or something. Default is 30s.~

~For 2, we have a setInterval that checks for updates every 30mins (?).~


src/context/importantMessage.tsx
```diff
@@ -0,0 +1,39 @@
...
+);
+export const ImportantMessageSetterContext = createContext<
+  Dispatch<SetStateAction<MessageContent | null>>
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+>(() => {});
+
+export const ImportantMessageContextProvider: FunctionComponent = ({
+  children
+}) => {
+  const [message, setMessage] = useState<MessageContent | null>(null);
```
yehjxraymond, line 30: would you want to consider making this a Banner component so that the code for the message context is not coupled with the Banner's props. 

Of course that is for the case of one single banner.. .

src/flags.ts
```diff
@@ -1,3 +1,4 @@
 export const Flags = {
-  HELP_MODAL: false
+  HELP_MODAL: false,
+  AUTO_CHECK_UPDATES: true
```
yehjxraymond, line 3: why add a flag which is set to true anyway?

sebastianquek: Yea, I can remove this for now. Was preparing for the use case of SA being used in a self-service kiosk of sorts 🙃

src/context/importantMessage.tsx
sebastianquek: Hmm, yea I'll look into this in another PR where I'll handle multiple messages better.

PR #25, opened 2020-04-07
feat: add InputNricSection

sebastianquek:
Extracting out this part so that we can use it in modals and other screens

image


src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
```diff
@@ -93,9 +61,7 @@ const CollectCustomerDetailsScreen: FunctionComponent<NavigationFocusInjectedPro
 
   const onCheck = async (input: string): Promise<void> => {
     try {
-      const isNricValid = validate(input);
-      if (!isNricValid) throw new Error("Invalid NRIC number");
-      const nric = input.match(nricRegex)?.[0].toUpperCase();
+      const nric = validateAndCleanNric(input);
 
       setIsLoading(true);
       const quota = await getQuota(nric!, authKey, endpoint);
```
yehjxraymond, line 67: No need `!` le right?

PR #26, opened 2020-04-07
feat: check quota in quota screen

sebastianquek:
Move checking of quota to `CustomerQuotaScreen` so we don't depend on `CollectCustomerDetailsScreen` to pass it through state (separation of concerns).

Also split `CustomerQuotaScreen` into smaller components as it was getting very long

### DIFF
--- a/src/components/CustomerQuota/NoQuotaCard.tsx
+++ b/src/components/CustomerQuota/NoQuotaCard.tsx
@@ -1,0 +1,80 @@
+import React, { FunctionComponent } from "react";
+import { Transaction } from "../../services/quota";
+import { differenceInSeconds, format, formatDistance } from "date-fns";
+import { View } from "react-native";
+import { CustomerCard } from "./CustomerCard";
+import { AppText } from "../Layout/AppText";
+import { color } from "../../common/styles";
+import { sharedStyles } from "./sharedStyles";
+import { DarkButton } from "../Layout/Buttons/DarkButton";
+
+const DURATION_THRESHOLD_SECONDS = 60 * 10; // 10 minutes
+
+interface NoQuotaCard {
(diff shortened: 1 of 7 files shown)


src/components/CustomerQuota/NoQuotaCard.tsx
```diff
@@ -0,0 +1,80 @@
...
+          style={[
+            sharedStyles.resultWrapper,
+            sharedStyles.failureResultWrapper
+          ]}
+        >
+          <AppText style={sharedStyles.emoji}>❌</AppText>
+          <AppText style={sharedStyles.statusTitleWrapper}>
+            {secondsFromLastTransaction > 0 ? (
+              secondsFromLastTransaction > DURATION_THRESHOLD_SECONDS ? (
+                <>
```
yehjxraymond, line 46: Was thinking if you want to put this in a separate component with names like:

RecentTransactionPanel
DistantTransactionPanel
NoPreviousTransactionPanel

sebastianquek: Yep sure!

src/components/CustomerQuota/types.ts
```diff
@@ -0,0 +1,3 @@
+export interface CartState {
+  [category: string]: boolean | null;
```
yehjxraymond, line 2: Come to think of it, do we want a custom type for this?

"Selected" | "NotSelected" | "NoQuota"

Can make the code a lot easier to read also

sebastianquek: Yep I'll be changing the way the cart works so that it's clearer. As we're moving to allow numeric on top of boolean values, this will change too.

PR #28, opened 2020-04-08
feat: support multiple NRICs

sebastianquek:
* Add necessary changes to support multiple NRICs (actually adding multiple NRICs will be handled in another PR)
* Add `useCart` hook that abstracts away handling of the cart
* Update quota service to use new endpoints

---
# NOTE
~Currently the `NoQuotaCard` cannot show the last transaction date since it's not returned by the API. Perhaps we can have another endpoint that returns the last transactions for each item for a particular NRIC. In that case, we can query whenever the `NoQuotaCard` needs to be shown.~

Will use the current `/quota/:id` endpoint when there's only 1 NRIC and this returns the transaction time. So will use this in `NoQuotaCard`

---
On testing hooks with context: https://github.com/testing-library/react-hooks-testing-library/blob/master/docs/usage/advanced-hooks.md

### DIFF
--- a/src/components/CustomerQuota/NoQuotaCard.tsx
+++ b/src/components/CustomerQuota/NoQuotaCard.tsx
@@ -1,41 +1,50 @@
 import React, { FunctionComponent } from "react";
-// import { Transaction } from "../../services/quota";
-// import { differenceInSeconds, format, formatDistance } from "date-fns";
-import { View } from "react-native";
+import { compareDesc } from "date-fns";
+import { differenceInSeconds, format, formatDistance } from "date-fns";
+import { View, StyleSheet } from "react-native";
 import { CustomerCard } from "./CustomerCard";
 import { AppText } from "../Layout/AppText";
-import { color } from "../../common/styles";
+import { color, size, fontSize } from "../../common/styles";
 import { sharedStyles } from "./sharedStyles";
 import { DarkButton } from "../Layout/Buttons/DarkButton";
 import { Cart } from "../../hooks/useCart/useCart";
+import { useProductContext } from "../../context/products";
 
-// const DURATION_THRESHOLD_SECONDS = 60 * 10; // 10 minutes
+const DURATION_THRESHOLD_SECONDS = 60 * 10; // 10 minutes
 
-// const DistantTransactionContent: FunctionComponent<{
-//   transactionTime: number;
-// }> = ({ transactionTime }) => (
-//   <>
-//     <AppText style={sharedStyles.statusTitle}>Limit reached on </AppText>
-//     <AppText style={sharedStyles.statusTitle}>
-//       {format(transactionTime, "hh:mm a, do MMMM")}.
-//     </AppText>
-//   </>
-// );
+const styles = StyleSheet.create({
+  itemTransactions: {
+    marginTop: size(1),
+    lineHeight: 1.5 * fontSize(0),
+    marginBottom: -size(2)
+  }
+});
 
-// const RecentTransactionContent: FunctionComponent<{
-//   now: Date;
-//   transactionTime: number;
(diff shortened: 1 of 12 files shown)


src/services/quota/index.tsx
```diff
@@ -1,87 +1,114 @@
...
-  transactionTime: number;
+  transactionTime?: number;
 }
 
 export interface Quota {
-  remainingQuota: Transaction[];
+  remainingQuota: QuotaItem[];
 }
 
 export interface PostTransaction {
```
ilhaoni, line 13: Should we rename this as GroupTransaction instead? Might be clearer haha

src/services/quota/index.tsx
```diff
@@ -1,87 +1,114 @@
...
   key: string;
   endpoint: string;
 }
 
+export interface TransactionItemResponse {
+  transaction: Pick<QuotaItem, "category" | "quantity">[];
+  timestamp: number;
+}
+
 export interface PostTransactionResponse {
```
ilhaoni, line 25: And this to GroupTransactionResponse

src/hooks/useCart/useCart.test.ts
```diff
@@ -0,0 +1,415 @@
...
+      expect(result.current.error?.message).toBe("Insufficient quota");
+      expect(result.current.cart).toStrictEqual({
+        "toilet-paper": {
+          category: "toilet-paper",
+          quantity: 0,
+          maxQuantity: 2
+        },
+        chocolate: {
+          category: "chocolate",
+          quantity: 0,
```
ilhaoni: hmm for this right, from a user point of view, if theres max 15, and i put a 100,

does it make more sense to have `quantity: 0` or `quantity: 15`? The insufficient quota message should still be thrown haha

sebastianquek: Hmm, I prefer to do nothing if there are invalid values though. i.e. nothing is "committed" to the cart unless it's valid

src/hooks/useCart/useCart.ts
```diff
@@ -0,0 +1,171 @@
...
+          nrics: ids,
+          key: authKey,
+          transactions,
+          endpoint
+        });
+        setCheckoutResult(transactionResponse);
+        setCartState("PURCHASED");
+      } catch (e) {
+        setError(new Error("Couldn't checkout, please try again later"));
+        setCartState("DEFAULT");
```
ilhaoni: ah so here, we'll have to get the cashiers to click "checkout" again so that the whole system will retry right

ilhaoni:
Ahhh, understand understand. I think that’s a good point! Let’s do nothing then

________________________________

src/hooks/useCart/useCart.ts
sebastianquek: Yea, in future we can add a library that helps us retry (:

ilhaoni:
Oooh, Okk! \^^/

________________________________

src/components/CustomerQuota/NoQuotaCard.tsx
```diff
@@ -69,21 +96,29 @@ export const NoQuotaCard: FunctionComponent<NoQuotaCard> = ({
...
                 />
               )
             ) : (
-              <NoPreviousTransactionContent />
+              <NoPreviousTransactionTitle />
             )}
           </AppText>
+          {itemTransactions.length > 0 && (
+            <View>
+              <AppText>When limits were reached:</AppText>
```
ilhaoni, line 116: This is nice haha

src/services/quota/index.tsx
sebastianquek: As discussed, we'll use the current approach since it may not be a group transaction. I've also added back `Transaction`

PR #32, opened 2020-04-11
[DO NOT MERGE] feat: otp flow

TangLiRen:
Implemented OTP flow with the following features:

- When user scans QR code, it allows them to type in their number
- Number is validated with a global number validation (can change to any country if we want to), and will show an alert if not valid
- When submitting number, will call `/auth/register` on backend. Error will show an alert in the frontend that they couldn't create the OTP, and will allow them to try again.
- On success user will enter OTP screen and will receive the sms, which they can key into the number field. There is a feature to resend OTP after 30sec (which resets upon resending). Upon resending, it disables the confirm button, and on confirming it disables the resend button. Errors on either will show alert on frontend.
- Upon OTP confirmation success, it will save the `sessionToken`, `endpoint` and `expiry` to the `AuthenticationContext` and `AsyncStorage` (we don't actually need it in `AuthenticationContext` but will leave it there to prevent breaking changes and the need for quite significant refactoring)
- User will be logged out automatically after expiry is over, and will be prompted 15 mins before the expiry is up.
- On logout prompt, user can choose to log out instantly or cancel the prompt and continue their work (in which they will be automatically logged out on expiry without further warning)
- There is now a log out button at the side of the `AppName` that opens a prompt to confirm logging  out.
- Upon successful log out there will be a prompt to tell users they have done so successfully and will always route them back to login page. clearing the `AuthenticationContext` and `AsyncStorage`.


src/components/Layout/AppHeader.tsx
```diff
@@ -0,0 +1,106 @@
...
+const TIME_BEFORE_WARNING = 900000;
+let warningTimer: NodeJS.Timeout;
+let logoutTimer: NodeJS.Timeout;
+
+interface AppHeader extends NavigationProps {
+  mode?: AppMode;
+}
+
+const styles = StyleSheet.create({
+  AppHeaderWrapper: {
```
sebastianquek, line 21: Should be camelCase

src/components/Layout/AppHeader.tsx
```diff
@@ -0,0 +1,106 @@
...
+  return (
+    <View style={styles.AppHeaderWrapper}>
+      <AppName mode={mode} />
+      <BaseButton onPress={onPressLogout} iconOnly>
+        <MaterialCommunityIcons
+          name="logout"
+          size={size(3)}
+          color={color("grey", 0)}
+        />
+      </BaseButton>
```
sebastianquek, line 101: I think we should show the text "Logout", not easy to understand that icon. We don't have a smaller variant of the buttons atm, I suggest to just style a special button for this page first.

src/components/Layout/Buttons/BaseButton.tsx
```diff
@@ -8,6 +8,7 @@ export interface BaseButton {
   borderColor?: ViewStyle["borderColor"];
   disabled?: boolean;
   fullWidth?: boolean;
+  iconOnly?: boolean;
```
sebastianquek, line 11: May not need this, unlikely we'll have icon only buttons since they aren't very recognisable/understandable

src/components/Login/LoginContainer.tsx
```diff
@@ -101,86 +106,81 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
...
-
   const onBarCodeScanned = (event: BarCodeScanningResult): void => {
     if (!isLoading && event.data) {
+      const qrCode = event.data;
       onToggleScanner();
-      onLogin(event.data);
+      setIsLoading(true);
+      const { key, endpoint } = decodeQr(qrCode);
+      setCodeKey(key);
+      setEndpointTemp(endpoint);
```
sebastianquek, line 116: Hmm, what's the reason for this being 'temp'

src/context/auth.tsx
```diff
@@ -2,21 +2,29 @@ import React, {
...
+import { AsyncStorage } from "react-native";
+
+export const SESSION_TOKEN_KEY = "SESSION_TOKEN";
+export const EXPIRY_KEY = "EXPIRY_KEY";
+export const ENDPOINT_KEY = "ENDPOINT_KEY";
 
 interface AuthenticationContext {
-  authKey: string;
+  token: string;
+  expiry: string;
```
sebastianquek, line 16: Would storing this as a number be better? So consumers don't need to convert yet again

src/services/auth/index.tsx
```diff
@@ -1,57 +1,69 @@
 import { IS_MOCK } from "../../config";
-import { Policy } from "../../types";
```
sebastianquek, line 2: We'll need to add this back for now

src/components/Login/LoginContainer.tsx
TangLiRen: Cause we can get the endpoint straight from the QR code, but I feel its better to only set it in the context and the async storage after confirming otp, so the component will store inside the local state as a "temp" storage

src/context/auth.tsx
TangLiRen: yeah true

src/services/auth/index.tsx
TangLiRen: as in the whole Policy being returned from `/confirm`?

sebastianquek: We'll use `/auth` first, and later refactor

PR #39, opened 2020-04-14
feat: error boundary + sentry

sebastianquek:
* Error boundary to handle all uncaught errors gracefully
* Add env for Sentry variables
* Add breadcrumbs for better context in Sentry when errors are thrown
* `LoginStage` is now a union type so the breadcrumb sent to Sentry is more descriptive

---

To test, open this on the published channel, enter `123` as phone number to show the error boundary screen


src/components/Login/LoginMobileNumberCard.tsx
```diff
@@ -79,6 +80,12 @@ export const LoginMobileNumberCard: FunctionComponent<LoginMobileNumberCard> = (
   const [countryCode, setCountryCode] = useState("+65");
   const [mobileNumberValue, setMobileNumberValue] = useState("");
 
+  useEffect(() => {
```
yehjxraymond, line 83: was thinking if you want to do a hook so that this chunk dont need to copy paste everywhere haha

src/types/modules.d.ts
```diff
@@ -3,3 +3,11 @@ declare module "*.svg" {
   const content: React.StatelessComponent<SvgProps>;
   export default content;
 }
+
+declare module "react-native-dotenv" {
+  export const SENTRY_ORG: string;
+  export const SENTRY_PROJECT: string;
+  export const SENTRY_AUTH_TOKEN: string;
+  export const SENTRY_DSN: string;
+  export const ENV: "dev" | "prod";
```
yehjxraymond, line 12: we are not using this right?

sebastianquek: oh ya, forgot to remove, thanks

src/components/Login/LoginMobileNumberCard.tsx
sebastianquek: Yea was thinking if it's possible, for the on mount navigation events it'll work, for the other kinds such as when `loginStage` changes, not sure if it's beneficial since I'll essentially be replicating the useEffect logic with the dependency array. Any idea how this can be done?

PR #42, opened 2020-04-20
feat: improve logout flow

sebastianquek:
- When app is freshly opened / became in focus
    - If token is expired, logout immediately
    - If token is about to expire (60mins), show banner
- When app has been kept open
    - Whenever an NRIC is scanned, if about to expire (60mins), show banner 
    - Once countdown reaches 0, logout immediately

| Countdown updates every minute | When less than a minute, changes every second | Automatic logout |
| --- | --- | --- |
|Simulator Screen Shot - iPhone 11 Pro - 2020-04-20 at 16 24 41| Simulator Screen Shot - iPhone 11 Pro - 2020-04-20 at 16 29 22 | Simulator Screen Shot - iPhone 11 Pro - 2020-04-20 at 16 30 15|

### DIFF
--- a/src/hooks/useLogout.tsx
+++ b/src/hooks/useLogout.tsx
@@ -1,0 +1,51 @@
+import { useContext, useState, useCallback } from "react";
+import { ImportantMessageSetterContext } from "../context/importantMessage";
+import { useAuthenticationContext } from "../context/auth";
+import { useProductContext } from "../context/products";
+import { Alert } from "react-native";
+import { NavigationDispatch, NavigationActions } from "react-navigation";
+
+type AlertProps = {
+  title: string;
+  description?: string;
+};
+
+interface LogoutHook {
+  isLoggingOut: boolean;
+  logout: (navigationDispatch?: NavigationDispatch, alert?: AlertProps) => void;
+}
+
+export const useLogout = (): LogoutHook => {
+  const setMessageContent = useContext(ImportantMessageSetterContext);
+  const { clearAuthInfo } = useAuthenticationContext();
+  const { setProducts } = useProductContext();
+  const [isLoggingOut, setIsLoggingOut] = useState(false);
+
+  const logout: LogoutHook["logout"] = useCallback(
+    async (navigationDispatch, alert) => {
+      if (!navigationDispatch) {
+        return;
+      }
+      setIsLoggingOut(true);
+      await clearAuthInfo();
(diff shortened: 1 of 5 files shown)


src/hooks/useValidateExpiry.tsx
```diff
@@ -0,0 +1,75 @@
...
+    clearTimeout(timeout);
+
+    if (expiry === "" || isLoggingOut) {
+      return;
+    }
+    const diffInSeconds = Math.max(
+      0,
+      differenceInSeconds(Number(expiry), Date.now())
+    );
+    if (diffInSeconds === 0) {
```
yehjxraymond, line 56:
```suggestion
    if (diffInSeconds === 0) {
```

might want to use less than or equal to

src/hooks/useValidateExpiry.tsx
```diff
@@ -0,0 +1,75 @@
...
+      onExpired();
+    } else if (diffInSeconds <= ABOUT_TO_EXPIRE_SECONDS) {
+      onAboutToExpire(diffInSeconds);
+      let durationInSeconds = MAX_INTERVAL_SECONDS;
+      if (diffInSeconds > MAX_INTERVAL_SECONDS) {
```
yehjxraymond, line 61: think if we do the above then no need the complex if statements liao and just query every 1s

src/hooks/useLogout.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { NavigationDispatch, NavigationActions } from "react-navigation";
+
+type AlertProps = {
+  title: string;
+  description?: string;
+};
+
+interface LogoutHook {
+  isLoggingOut: boolean;
+  logout: (navigationDispatch?: NavigationDispatch, alert?: AlertProps) => void;
```
yehjxraymond, line 15: will there be a case where the dispatch is null?

src/hooks/useValidateExpiry.tsx
sebastianquek: Good catch, I don't need the max before this too.

src/hooks/useValidateExpiry.tsx
sebastianquek: I think better to not rerender every second once there's 60mins left, no real need also. I've added some comments in case this part wasn't very clear.

PR #45, opened 2020-04-23
feat: io-ts to validate and transform API responses

sebastianquek:
### Validate and decode objects
```ts
import * as t from "io-ts";
import { isRight } from "fp-ts/lib/Either";
import { DateFromNumber } from "io-ts-types/lib/DateFromNumber";

const SessionCredentials = t.type({
  sessionToken: t.string,
  ttl: DateFromNumber
});

type SessionCredentials = t.TypeOf<typeof SessionCredentials>;
// equivalent to
// type SessionCredentials = {
//    sessionToken: string;
//    ttl: Date;
// }

const decoded = SessionCredentials.decode({
  sessionToken: "token",
  ttl: 1587630596418  // endpoint returns a number that will be transformed to a Date 
});

if (isRight(decoded)) {
  console.log(decoded.right);
} else {
  // error decoding
}
```
Output:
```
Object {
  "sessionToken": "token",
  "ttl": 2020-04-23T08:29:56.418Z,  // Note that the number is converted to a date
}
```

### Optional types
There's no inbuilt `t.optional` and we need to rely on using `t.partial`:
```ts
const ItemQuota = t.intersection([
  t.type({
    category: t.string,
    quantity: t.number
  }),
  t.partial({
    transactionTime: DateFromNumber
  })
]);
type ItemQuota = t.TypeOf<typeof ItemQuota>;
```
This is equivalent to
```
type ItemQuota = {
    category: string;
    quantity: number;
} & {
    transactionTime?: Date | undefined;
}
```

We could use `t.union([ t.undefined, DateFromNumber ])`, but this would mean that `transactionTime` will need to be explicitly defined as `undefined` whenever we want to omit it.

### Usage when fetching from an endpoint
If the API response is ok, it'll decode. `report(decoded)` will print out the exact fields that are mismatched. In our case, we'll let Sentry know if a validation error occured since this implies something wrong with our API endpoints.
```ts
// services/helpers.ts
import { fold } from "fp-ts/lib/Either";
import { Type } from "io-ts";
import { reporter } from "io-ts-reporters";

export async function fetchWithValidator<T, O, I>(
  validator: Type<T, O, I>,
  requestInfo: RequestInfo,
  init?: RequestInit
): Promise<T> {
  const response = await fetch(requestInfo, init);

  const json = await response.json();
  if (!response.ok) {
    throw new Error(json.message ?? "Fetch failed");
  }

  const decoded = validator.decode(json);
  return fold( 
    () => {
      throw new Error(reporter(decoded).join(" "));
    },
    (value: T) => value
  )(decoded);
}

```


src/hooks/useCart/useCart.test.tsx
```diff
@@ -97,37 +109,37 @@ const mockQuotaResMultipleIds = {
...
     }
   ]
 };
 
 describe("useCart", () => {
   beforeEach(() => {
-    mockFetch.mockReset();
+    mockGetQuota.mockReset();
+    mockGetPolicies.mockReset();
+    mockPostTransaction.mockReset();
```
ilhaoni, line 131: Just curious, would `jest.resetAllMocks()` be simpler here?

sebastianquek: Yep, I've updated it! I think just need to be mindful about the global mocks that we have, but in this case it's fine!

ilhaoni: Ohh ya, that's true haha. Okk~

PR #49, opened 2020-06-03
feat: successful CDC flow

szenius:
[Trello card](https://trello.com/c/wtJhbXEM/72-cdc-flow-voucher-redemption) | [Figma screens](https://www.figma.com/proto/vYCze8TIVyrbvuF7IZliRP/SupplyAlly?node-id=628%3A5665&scaling=min-zoom)

- [x] Added new item selection view for policies with identifiers
- [x] All identifiers must be filled for checkout to not throw error
- [x] Added Redeemed! screen for policies with type "redeem"
  * Only checks the first policy, if it is a "redeem" type, assume all policies are all also "redeem" type
- [x] Add tests

# Testing Instructions
## Testing with backend
Can use the environment for https://github.com/rationally-app/backend-api/pull/26 to test!

## Testing with mocks
In `src/services/policies/index.ts`, change `mockGetPolicies` function to this
```
const mockGetPolicies = async (
  _token: string,
  _endpoint: string
): Promise<Policies> => {
  return {
    policies: [
      {
        category: "vouchers",
        name: "Vouchers",
        order: 1,
        identifiers: [
          { scannerType: "QR", label: "First" },
          { scannerType: "QR", label: "Last" }
        ],
        quantity: {
          period: 7,
          limit: 1,
          default: 1
        },
        type: "redeem"
      }
    ]
  };
};
```

Then in `src/services/quota/index.ts`, change `mockGetQuota` function to this
```
export const mockGetQuota = async (
  ids: string[],
  _key: string,
  _endpoint: string
): Promise<Quota> => {
  if (ids[0] === "S0000000J") throw new Error("Something broke");
  if (ids.length === 1) {
    const transactionTime = new Date(2020, 3, 5);
    return {
      remainingQuota: [
        {
          category: "vouchers",
          quantity: 1,
          transactionTime
        }
      ]
    };
  } else {
    return {
      remainingQuota: [
        {
          category: "vouchers",
          quantity: 1
        }
      ]
    };
  }
};
```

Run `npm start:mock` to test the CDC success flow

# TODO in a future MR
- Remove "Add" NRIC button in item selection view for policies with identifiers
- Add validation for identifiers
- Add"Limit reached" CDC flow


src/components/CustomerQuota/ItemsSelection/ItemIdentifiersCard.tsx
```diff
@@ -0,0 +1,48 @@
...
+      ...identifierInputs.slice(0, index),
+      { ...identifierInputs[index], value },
+      ...identifierInputs.slice(index + 1)
+    ]);
+  };
+
+  useEffect(() => {
+    updateCart(cartItem.category, cartItem.quantity, identifierInputs);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [identifierInputs]);
```
szenius, line 34: Including all deps was causing state to change too fast, so checking out became unpredictable:
* Sometimes the identifiers were not recorded when checking out
* Sometimes can't check out even after filling in the identifiers

Not sure if there is a better way to solve this!! What do you think?

ilhaoni: oh, serious? hmm, so adding the lint fixes it ah? :O

ilhaoni:
Hmm, im wondering if the change of mock values mentioned in this PR should be updated in our mocks as well, but I dont see any mocks so far

src/components/CustomerQuota/ItemsSelection/ItemIdentifiersCard.tsx
szenius: Lint keep saying need to add `cartItem.X` dependencies in too, but doesn't make sense cuz they won't change 😆  Adding lint avoids the warning so we can only add `identifierInputs` to deps array

szenius:
> Hmm, im wondering if the change of mock values mentioned in this PR should be updated in our mocks as well, but I dont see any mocks so far

Actually the mocks depend on what we want to test for that PR, for different policy it will be different. That's why I wrote in the description in case we want to test with mocks 😅

src/components/CustomerQuota/ItemsSelection/ItemIdentifiersCard.tsx
ilhaoni: ohh, i see i see! hmm yeah, maybe we can think about it, to see if there's a better solution. I need to think through also HAHA

ilhaoni:
> > Hmm, im wondering if the change of mock values mentioned in this PR should be updated in our mocks as well, but I dont see any mocks so far
> 
> Actually the mocks depend on what we want to test for that PR, for different policy it will be different. That's why I wrote in the description in case we want to test with mocks 😅

Thats true..... Ok i understand now! Yeah, maybe let me think about this more haha

PR #55, opened 2020-06-12
feat: CDC flow for ineligible transactions

dalsontws:
- [x] Screen for not eligible NRICs
- [x] unit tests

integration with mystique

- [x] throw EligibilityError for specific error message
- [x] catch EligibilityError and set cart state to Not Eligible

Note: 
Catch EligibilityError and set cart state is used ahead of checking the quotaResponse to determine error, as response is `null`, and could be due to other reasons


ilhaoni:
Currently we haven’t integrate with mystique to try yet right haha

Should we update the useCart.test as well? Since we now have new states ((:

src/hooks/useCart/useCart.tsx
```diff
@@ -131,8 +137,10 @@ export const useCart = (
         }
         setQuotaResponse(quotaResponse);
       } catch (e) {
-        // Cart will remain in FETCHING_QUOTA state.
-        if (e instanceof PolicyError) {
+        if (e instanceof EligibilityError) {
```
szenius, line 140: Nit, would it be more straight forward to call this `NotEligibleError`?

szenius, line 140: Nit, would it be more straight forward to call this NotEligibleError?

src/hooks/useCart/useCart.tsx
```diff
@@ -131,8 +137,10 @@ export const useCart = (
...
+          setCartState("NOT_ELIGIBLE");
```
szenius, line 141: Need to write tests~~

src/hooks/useCart/useCart.tsx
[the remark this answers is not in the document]
dalsontws: yeah wont! haha the error message needs to be specific to throw eligibilityError hehe

dalsontws: agreed haha!!👌

src/hooks/useCart/useCart.test.tsx
```diff
@@ -159,6 +164,13 @@ const mockPostTransactionResult = {
   ]
 };
 
+const mockQuotaNotEligible: any = (id: string) => {
```
szenius, line 167: Small nit too: might be more understandable if the quota word was replaced here, cuz it's not really that the quota that was not eligible

dalsontws: got it, changed! :)

src/components/CustomerQuota/IneligibleCard.tsx
```diff
@@ -0,0 +1,54 @@
...
+  return (
+    <View>
+      <CustomerCard nrics={nrics} headerBackgroundColor={color("red", 60)}>
+        <View
+          style={[
+            sharedStyles.resultWrapper,
+            sharedStyles.failureResultWrapper
+          ]}
+        >
+          <Text style={sharedStyles.emoji}>❌</Text>
```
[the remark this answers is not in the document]
dalsontws: agreee! adding this to new card, to discuss with Eida as well 👍

PR #57, opened 2020-06-13
fix: block checkout without voucher codes

szenius:
[Trello card](https://trello.com/c/FOWlOnLO/119-bug-checkout-can-happen-without-voucher-code-input)

This MR fixes the bug where sometimes users could checkout without inputting voucher codes. See Trello card for steps to reproduce. 

Issue was in mergeWithCart, where the identifier inputs were being overwritten.

Also did a bit of refactoring for unique code validation because I think we were unnecessarily validating N times when we can do it only 1 time.

Also made identifiers an optional parameter for updateCart, because it was difficult to debug where we were accidentally replacing identifier inputs with an empty array.

### DIFF
--- a/src/hooks/useCart/useCart.tsx
+++ b/src/hooks/useCart/useCart.tsx
@@ -221,7 +221,7 @@ export const useCart = (
             numUnverifiedTransactions += 1;
           } else {
             identiferValues.push(
-              ...identifiers.map(idenfitier => idenfitier.value)
+              ...identifiers.map(identifier => identifier.value)
             );
           }
 
--- a/src/hooks/useCart/useCart.tsx
+++ b/src/hooks/useCart/useCart.tsx
@@ -210,29 +210,26 @@ export const useCart = (
 
       let numUnverifiedTransactions = 0;
       let numIdentifiers = 0;
-      const identifierInputValues: string[] = [];
+      const identiferValues: string[] = [];
       const transactions = Object.values(cart)
         .filter(({ quantity }) => quantity)
-        .map(({ category, quantity, identifiers: identifierInputs }) => {
+        .map(({ category, quantity, identifiers }) => {
           if (
-            identifierInputs.length > 0 &&
-            identifierInputs.some(identifier => !identifier.value)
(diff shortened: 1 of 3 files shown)


src/hooks/useCart/useCart.tsx
```diff
@@ -210,29 +210,26 @@ export const useCart = (
 
       let numUnverifiedTransactions = 0;
       let numIdentifiers = 0;
-      const identifierInputValues: string[] = [];
+      const identiferValues: string[] = [];
       const transactions = Object.values(cart)
         .filter(({ quantity }) => quantity)
-        .map(({ category, quantity, identifiers: identifierInputs }) => {
+        .map(({ category, quantity, identifiers }) => {
```
dalsontws, line 216: Okay looks good to me! my initial idea was taking into consideration the possibility of doing a check for each category as well (i.e. Last voucher code > First voucher code), but not doing that liao haha

szenius: I see!! Actually if we wanted to do per category, we can just keep a boolean, no need to keep both boolean and array too

dalsontws: thats true!! hmm but array still needed to check uniqueness across categories right?

szenius: Can use the boolean!! Very simplified example:

```
let isUnique = true
for item in cart:
  isUnique = isUnique && checkIsUnique(item.identifiers)
```

src/hooks/useCart/useCart.tsx
```diff
@@ -215,19 +219,17 @@ export const useCart = (
             identifiers.some(identifier => !identifier.value)
           ) {
             numUnverifiedTransactions += 1;
+          } else {
+            identiferValues.push(
+              ...identifiers.map(idenfitier => idenfitier.value)
```
dalsontws, line 224: minor change! identifier spelling :)

szenius: HAHAHA wah good catch, I keep spelling it wrong omg

PR #58, opened 2020-06-13
Increment version number to publish App/Play Store

younglim:
Increment the version number for APK to be uploaded to Play Store.


sebastianquek:
Noticed this will be outdated by the time we want this update again, closing for now.

PR #60, opened 2020-06-16
chore: update policy identifiers

szenius:
This PR
* Updates policy identifiers to new shape
* Renames quota and transaction identifiers as identifierInputs for better differentiation
* Adds types to identifier inputs

(Test against [pr33](https://github.com/rationally-app/backend-api/pull/33))

### DIFF
--- a/src/hooks/useCart/useCart.test.tsx
+++ b/src/hooks/useCart/useCart.test.tsx
@@ -3,7 +3,12 @@ import { renderHook } from "@testing-library/react-hooks";
 import { useCart } from "./useCart";
 import { wait } from "@testing-library/react-native";
 import { ProductContext } from "../../context/products";
-import { Policy } from "../../types";
+import {
+  Policy,
+  Quota,
+  PostTransactionResult,
+  PolicyIdentifier
+} from "../../types";
 import {
   getQuota,
   postTransaction,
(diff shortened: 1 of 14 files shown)


src/services/policies/index.ts
```diff
@@ -78,6 +78,24 @@ const mockGetPolicies = async (
...
+        order: 4,
+        quantity: { period: 1, limit: 1 },
+        identifiers: [
+          {
+            label: "Voucher code",
+            textInput: { visible: true, disabled: true, type: "STRING" },
+            scanButton: {
+              visible: true,
+              disabled: false,
+              type: "QR",
```
ilhaoni, line 94: actually, should we keep it as `type: "BARCODE"` haha

src/hooks/useCart/useCart.test.tsx
```diff
@@ -38,8 +49,14 @@ const defaultProducts: Policy[] = [
       }
     },
     identifiers: [
-      { scannerType: "QR", label: "first" },
-      { scannerType: "QR", label: "last" }
+      {
+        ...defaultIdentifier,
+        label: "first"
```
ilhaoni, line 54: ehh, i'm a bit confused here, so there's 2 labels ah, one inside the defaultIdentifier, and one more outside? :O

src/services/policies/index.ts
szenius: Oh right, can also~~ Not much impact now cuz it's not supported anyway HAHA

ilhaoni: HAHA, its okk~~ at least we make it accurate 😂  cause now we'll send the type into the db also haha

szenius: Ohhh, this one won't get sent into the db. It only gets activated when we run `npm run start:mock`, which mocks the backend~

PR #61, opened 2020-06-16
feat: new endpoint to fetch feature flags

dalsontws:
using the new endpoint `/version` instead of `/auth` to fetch policies and features

- [x] using new endpoint `/version` instead of `/auth` [trello](https://trello.com/c/h9lJVRHV)
- [x] using `TRANSACTION_GROUPING` variable from feature to toggle "+add" button [trello](https://trello.com/c/6Ubf6h2e)
- [x] unit tests

### To test:
(locally) 
go to `src/services/policies/index.ts`

```
features: {
      REQUIRE_OTP: true,
      TRANSACTION_GROUPING: true
    }
```
toggle the `TRANSACTION_GROUPING` between true and false.
true will remove the add button from the CustomerCard

for PRs, toggle `TRANSACTION_GROUPING` in AWS Systems Manager


src/types.ts
```diff
@@ -50,14 +50,21 @@ const Policy = t.intersection([
   })
 ]);
 
+const Feature = t.type({
+  REQUIRE_OTP: t.boolean,
+  TRANSACTION_GROUPING: t.boolean
+});
+
 export const Policies = t.type({
```
ilhaoni, line 58: do you think it'll be cleaner if we rename the overall umbrella of `Policies` to something like
- `StageVersion`
- `EnvVersion`
- `DistVersion`

Ideally, stuff like env/distribution name and build number can come under this also haha. Cause it's meant to be version of this app haha.

What do yall think? ^^

src/types.ts
```diff
@@ -50,14 +50,21 @@ const Policy = t.intersection([
...
-  policies: t.array(Policy)
+  policies: t.array(Policy),
+  features: Feature
```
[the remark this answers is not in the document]
ilhaoni: but i would say this is of lower priority? so we can skip this also haha.

but renaming the overall `Policies` would be higher priority HAHA :D

src/types.ts
ilhaoni: cause if not currently its `Policies.policies` and `Policies.features`, and the features part dont really make sense haha

dalsontws: yes agree! i was quite confused with Policies having policies and features as well!

ilhaoni: haha yeahh~ which naming should we pick then haha

dalsontws: im fine with `DistVersion` or `EnvVersion` !

ilhaoni: okk~ how bout `EnvVersion` then? haha

@szenius you leh HAHA

src/hooks/useCart/useCart.tsx
```diff
@@ -151,7 +157,7 @@ export const useCart = (
         if (e instanceof NotEligibleError) {
           setCartState("NOT_ELIGIBLE");
           // Cart will remain in FETCHING_QUOTA state.
-        } else if (e instanceof PolicyError) {
+        } else if (e instanceof EnvVersionError) {
           setError(
             new Error(
               "Encountered an issue obtaining policies. We've noted this down and are looking into it!"
```
ilhaoni, line 163: small change here, shall we change the wording to `obtaining environment information`, or `obtaining version info`?

dalsontws: okie!

PR #67, opened 2020-06-17
Feat/merchant check validity modal

dhemutton:
This PR includes:
- Added 3 states to voucher checking
- Added invalid voucher modal
- Added loading modal

Trello link:
https://trello.com/c/DlFGTZcT


src/components/MerchantPayout/MerchantPayoutScreen.tsx
```diff
@@ -78,42 +82,52 @@ export const MerchantPayoutScreen: FunctionComponent<NavigationFocusInjectedProp
...
 
   const onBarCodeScanned: BarCodeScannedCallback = event => {
     if (isFocused && isScanningEnabled && event.data) {
       onCheckVoucher(event.data);
+    } else {
+      setVoucherStatus({
+        status: "INVALID",
+        errorMessage: "Please try Scanning again or enter voucher manually",
+        errorTitle: "Error scanning"
+      });
```
sebastianquek, line 130: Any validation should be done within the `onCheckVoucher`

src/components/MerchantPayout/VoucherStatusModal/InvalidCard.tsx
```diff
@@ -0,0 +1,55 @@
...
+  headerText: {
+    fontFamily: "brand-bold",
+    fontSize: fontSize(3),
+    marginBottom: size(2.5)
+  }
+});
+
+interface InvalidCard {
+  title: string;
+  details?: string;
```
sebastianquek, line 30: Are you thinking of a case when we wont show details?

src/components/MerchantPayout/VoucherStatusModal/LoadingCard.tsx
```diff
@@ -0,0 +1,23 @@
...
+  }
+});
+
+export const LoadingCard: FunctionComponent = () => {
+  return (
+    <Card style={[styles.card]}>
+      <LoadingView />
+    </Card>
+  );
+};
```
sebastianquek, line 23: Forgot to mention, the loading screen will just be an ActivityIndicator, so no need a card

src/components/MerchantPayout/VoucherStatusModal/VoucherStatusModal.tsx
```diff
@@ -0,0 +1,72 @@
...
+  return (
+    <Modal
+      visible={isVisible}
+      onRequestClose={onExit}
+      animationType="fade"
+      transparent={true}
+    >
+      <TouchableWithoutFeedback onPress={onExit}>
+        <View style={styles.background} />
+      </TouchableWithoutFeedback>
```
sebastianquek, line 68: Will be better if when there's an error, the user can only continue by tapping on the CTA, rather than anywhere outside. I think in this case, there isn't a scenario where we want this functionality

src/components/MerchantPayout/VoucherStatusModal/VoucherStatusModal.tsx
```diff
@@ -0,0 +1,72 @@
...
+  },
+  cardWrapper: {
+    padding: size(3),
+    flex: 1,
+    alignItems: "center",
+    justifyContent: "center"
+  }
+});
+
+type VoucherStatuses = "CHECKING" | "INVALID" | "VALID";
```
sebastianquek, line 26: One way we can improve testability and encapsulation is to move the checking of voucher status into our own custom hook. With the hook, any consumer can call `checkValidity` can be updated on the `checkValidityStatus` and the eventual error or success result. The consumer can listen in on any of the changes and react accordingly. This will allow us to test that specific hook on its own without relying on the UI portions. Let me know if you're unsure.

src/components/MerchantPayout/MerchantPayoutScreen.tsx
dhemutton: Ohh, we weren't sure if we had to have an error scanning modal based on the figma design and the else statement was to reflect that. The `onCheckVoucher` is called only if something was scanned successfully (but not validated). But we can remove the else statement too if u want!
Screenshot 2020-06-18 at 10 08 11 PM

src/components/MerchantPayout/VoucherStatusModal/InvalidCard.tsx
dhemutton: Yep, we were unsure if there will be other details in the future but we will amend it!

src/components/MerchantPayout/VoucherStatusModal/LoadingCard.tsx
dhemutton: So a greyed out background with no card? or just replace the LoadingView with the ActivityIndicator?

sebastianquek: Yes greyed out background, no card, white ActivityIndicator in the middle. You can also update LoadingView to include a color prop.

src/components/MerchantPayout/MerchantPayoutScreen.tsx
dhemutton: ahh i see...currently our catch error will result in the first modal (Invalid status header) with the error message as either "Invalid voucher code" or "Duplicate voucher code". Would you like us to differentiate the errors and show the second modal (Error scanning header)?

src/components/MerchantPayout/VoucherStatusModal/VoucherStatusModal.tsx
dhemutton: Okay! So should the `VoucherStatus {
  status: VoucherStatuses;
  errorMessage?: string;
  errorTitle?: string;
}` be in the hook as well or just `VoucherState = "CHECKING" | "INVALID" | "VALID";`? Also, will the validation of the voucher (from util) be called in the hook or on the UI side?

sebastianquek: The validation of the voucher from util can be called in the hook too!

As a consumer of this hook, I want to know the following:
```
type CheckValidityHook {
  checkValidityState: "DEFAULT" | "VALIDATING_ID" | "CHECKING_VALIDITY" | "RESULT_RETURNED";
  checkValidity: () => void;
  result: unknown; // need to figure out the type of this
  error?: Error;
  reset: () => void;
}
```

Then the consumer can call `checkValidity()` and react to any of these changes as they please.

This is just one way of looking at it!! I've used it and it works well, but feel free to explore!

PR #68, opened 2020-06-17
feat: text input UI for identifiers v2

szenius:
[Trello card](https://trello.com/c/SvdJlgy9/87-ui-tt-token-quota-screen)

This PR adds
* Support for `visible` and `disabled` attributes for text/scan inputs for identifiers
* Support for `STRING`/`NUMBER`/`PHONE_NUMBER` for identifiers' text input type
* Validation of phone numbers before checking out
* Hide identifier input display on Checkout screen if less than 2 identifiers
* Minor copy changes

Assumptions:
* We will never have an identifier with `textInput.type === "PHONE_NUMBER"` and scan button is also visible/enabled

Will add support for different scanner types in a different PR so this doesn't become too large!

### DIFF
--- a/src/components/CustomerQuota/CheckoutSuccess/RedeemedItem.tsx
+++ b/src/components/CustomerQuota/CheckoutSuccess/RedeemedItem.tsx
@@ -18,7 +18,7 @@ export const RedeemedItem: FunctionComponent<{
       <View style={sharedStyles.itemRow}>
         <AppText style={sharedStyles.itemHeaderText}>{categoryName}</AppText>
       </View>
-      {identifierInputs && identifierInputs.length > 0 && (
+      {identifierInputs && identifierInputs.length > 1 && (
         <View style={sharedStyles.quantitiesWrapper}>
           <View style={sharedStyles.quantitiesBorder} />
           <AppText style={sharedStyles.quantityByIdText}>
--- a/src/components/CustomerQuota/CheckoutSuccess/RedeemedItem.tsx
+++ b/src/components/CustomerQuota/CheckoutSuccess/RedeemedItem.tsx
@@ -13,18 +13,28 @@ export const RedeemedItem: FunctionComponent<{
   const { getProduct } = useProductContext();
   const { category, identifierInputs } = itemQuantities;
   const categoryName = getProduct(category)?.name ?? category;
+
+  let identifierInputDisplay = "";
+  if (identifierInputs) {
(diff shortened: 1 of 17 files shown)


src/components/CustomerQuota/ItemsSelection/ItemIdentifier.tsx
```diff
@@ -82,39 +127,36 @@ export const ItemIdentifier: FunctionComponent<{
...
-                <Feather
-                  name="maximize"
-                  size={size(2)}
-                  color={color("grey", 0)}
-                />
-              }
-              disabled={scanButton.disabled}
-              fullWidth={!textInput.visible}
-              onPress={() => setShouldShowCamera(true)}
+        {textInput.visible &&
```
ilhaoni, line 133: 🎊

src/components/CustomerQuota/CheckoutSuccess/RedeemedItem.tsx
```diff
@@ -18,7 +18,7 @@ export const RedeemedItem: FunctionComponent<{
       <View style={sharedStyles.itemRow}>
         <AppText style={sharedStyles.itemHeaderText}>{categoryName}</AppText>
       </View>
-      {identifierInputs && identifierInputs.length > 0 && (
+      {identifierInputs && identifierInputs.length > 1 && (
```
ilhaoni, line 21: what if there's only 1 input though? :x

szenius: According to the TT screens, no need to display, so I just use this hahahaha.

Hmmmm but for TT are we doing 2 identifiers per item ah? If yes then this will be a problem~

szenius: Maybe I can do this check

```
show identifier values if:
  identifierInputs.length > 0 && identifierInputs[0].type === identifierInputs[1].type
```

?

szenius: Will change this to display the single identifier when there is only one, and a range when more than 2 for now?

ilhaoni: okk! sounds good~ ((:

szenius: Updated :D

PR #70, opened 2020-06-18
style: shorten id input label

dalsontws:
[trello](https://trello.com/c/m2p4SHmr)

Currently "Enter Identification number" label wraps, quick fix to shorten label.
"Enter ID number" working for edge case (Nexus One)


ilhaoni:
> Probably outside the scope of this PR but why aren't there tests to update... Does that mean we have no tests for `validateAndCleanNric`???

HAHA i went to check. Ya we dont, we only have tests for validate :x 

We should write tests right HAHA. But not in this PR HAHA

PR #78, opened 2020-06-23
feat: mask phone numbers in no quota and redeem success screens

szenius:
This PR
* Adds new util function to display identifier inputs, which also masks identifier inputs when they are phone numbers
* Above util function is now used for no quota and redeemed screens

### DIFF
--- a/src/utils/getIdentifierInputDisplay.tsx
+++ b/src/utils/getIdentifierInputDisplay.tsx
@@ -1,0 +1,31 @@
+import { IdentifierInput } from "../types";
+
+const processIdentifierInputValue = (
+  identifierInput: IdentifierInput
+): string => {
+  return identifierInput.textInputType === "PHONE_NUMBER"
+    ? `****${identifierInput.value.slice(identifierInput.value.length - 4)}`
+    : identifierInput.value;
+};
+
+export const getIdentifierInputDisplay = (
+  identifierInputs: IdentifierInput[]
+): string => {
+  const filteredInputs = identifierInputs.filter(
+    identifierInput => !!identifierInput.value
+  );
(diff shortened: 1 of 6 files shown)


src/utils/getIdentifierInputDisplay.tsx
```diff
@@ -0,0 +1,31 @@
+import { IdentifierInput } from "../types";
+
+const processIdentifierInputValue = (
+  identifierInput: IdentifierInput
+): string => {
+  return identifierInput.textInputType === "PHONE_NUMBER"
+    ? `****${identifierInput.value.slice(identifierInput.value.length - 4)}`
```
sebastianquek, line 7: Perhaps can use `padStart`?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart

szenius: From discussion with Pat, we can display fixed 4 *s at the start. Let me ask if she prefers to pad as per original length, if yes I'll change to this. If not, I think might be more straightforward to do 4 *s?

szenius:
@ilhaoni Good catch, made the fix!

src/utils/getIdentifierInputDisplay.tsx
sebastianquek: Oh I was thinking to move this to some generic `maskText` function that can be used for anything, including NRIC, so just need to configure how many characters to show.

szenius: Pat said prefers to retain length, so I changed to padStart. Thanks for the suggestion 🙌

PR #79, opened 2020-06-23
feat: use of feature flag to toggle between environments

dalsontws:
[trello](https://trello.com/c/4rkGvfNe)

Feature flags previously called with `getQuota`. 
Should be called early on (after OTP), allowing toggling between environments/distributions

Adding additional feature variable to toggle.

```
      REQUIRE_OTP: true,
      TRANSACTION_GROUPING: true,
      DIST_ENV: "MERCHANT" / "VOUCHER" / "TOKEN"
```

String used instead of multiple boolean flags to prevent too many checks, possible issues with multiple environments resolving to `true`

Approach for toggling between Voucher collection and Merchant:
- immediately after validating OTP, `DIST_ENV` flag will be set, if flag is invalid, reset login state and return to initial login screen
- add new merchant flow screens into `StackNavigator`  
- use flag to determine navigation between different environments/distributions/flows

### DIFF
--- a/src/components/Login/LoginOTPCard.tsx
+++ b/src/components/Login/LoginOTPCard.tsx
@@ -82,9 +82,9 @@ export const LoginOTPCard: FunctionComponent<LoginOTPCard> = ({
       setProducts(versionResponse.policies);
 
       // Toggle between different environments
-      // using the TOGGLE_ENV variable from features
+      // using the DIST_ENV variable from features
 
-      // versionResponse.features.TOGGLE_ENV
+      // versionResponse.features.DIST_ENV
 
       navigation.navigate("CollectCustomerDetailsScreen");
     } catch (e) {
--- a/src/components/Login/LoginOTPCard.tsx
+++ b/src/components/Login/LoginOTPCard.tsx
@@ -86,7 +86,12 @@ export const LoginOTPCard: FunctionComponent<LoginOTPCard> = ({
 
       // versionResponse.features.DIST_ENV
 
-      navigation.navigate("CollectCustomerDetailsScreen");
+      if (versionResponse.features.DIST_ENV === "VOUCHER") {
+        navigation.navigate("CollectCustomerDetailsScreen");
+      } else {
+        alert("Navigation error");
+        navigation.navigate("loginStage");
+      }
     } catch (e) {
       if (e instanceof EnvVersionError) {
         alert(
--- a/src/components/Login/LoginOTPCard.tsx
+++ b/src/components/Login/LoginOTPCard.tsx
@@ -1,4 +1,9 @@
-import React, { useState, useEffect, FunctionComponent } from "react";
+import React, {
+  useState,
+  useEffect,
+  FunctionComponent,
+  useCallback
+} from "react";
 import { View, StyleSheet } from "react-native";
 import { DarkButton } from "../Layout/Buttons/DarkButton";
 import { SecondaryButton } from "../Layout/Buttons/SecondaryButton";
@@ -11,6 +16,7 @@ import { useAuthenticationContext } from "../../context/auth";
 import { validateOTP, requestOTP } from "../../services/auth";
 import { getEnvVersion, EnvVersionError } from "../../services/envVersion";
 import { useProductContext } from "../../context/products";
+import { useLogout } from "../../hooks/useLogout";
 
 const RESEND_OTP_TIME_LIMIT = 30 * 1000;
 
@@ -50,6 +56,11 @@ export const LoginOTPCard: FunctionComponent<LoginOTPCard> = ({
   );
   const { setAuthInfo } = useAuthenticationContext();
   const { setFeatures, setProducts } = useProductContext();
+  const { logout } = useLogout();
+
+  const handleLogout = useCallback((): void => {
+    logout(navigation.dispatch);
+  }, [logout, navigation.dispatch]);
 
   useEffect(() => {
     const resendTimer = setTimeout(() => {
@@ -90,7 +101,7 @@ export const LoginOTPCard: FunctionComponent<LoginOTPCard> = ({
         navigation.navigate("CollectCustomerDetailsScreen");
       } else {
         alert("Navigation error");
-        navigation.navigate("loginStage");
+        handleLogout();
       }
     } catch (e) {
       if (e instanceof EnvVersionError) {
--- a/src/components/Login/LoginOTPCard.tsx
+++ b/src/components/Login/LoginOTPCard.tsx
@@ -39,14 +39,14 @@ const styles = StyleSheet.create({
 });
 
 interface LoginOTPCard extends NavigationProps {
-  setLoginStage: Dispatch<SetStateAction<LoginStage>>;
+  resetStage: () => void;
   mobileNumber: string;
   codeKey: string;
   endpoint: string;
 }
 
 export const LoginOTPCard: FunctionComponent<LoginOTPCard> = ({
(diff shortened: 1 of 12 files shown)


src/hooks/useLogout.tsx
```diff
@@ -29,6 +29,11 @@ export const useLogout = (): LogoutHook => {
       setIsLoggingOut(true);
       await clearAuthInfo();
       setProducts([]);
+      setFeatures({
+        REQUIRE_OTP: true,
+        TRANSACTION_GROUPING: true,
+        DIST_ENV: ""
+      });
```
sebastianquek, line 36: Since we're always populating this from `/version` call, do we need to specify these values? Can we just have it be an empty object?

src/components/Login/LoginOTPCard.tsx
```diff
@@ -69,11 +81,41 @@ export const LoginOTPCard: FunctionComponent<LoginOTPCard> = ({
...
+        setProducts(versionResponse.policies);
+
+        navigation.navigate("CollectCustomerDetailsScreen");
+      } else {
+        alert(
+          "Invalid Environment Error: Make sure you scanned a valid QR code"
+        );
+        // Reset to initial login state
+        setLoginStage("SCAN");
+      }
```
sebastianquek, line 109: Shall we use a `switch` here so we ensure every case is taken care of

src/components/Login/LoginOTPCard.tsx
```diff
@@ -69,11 +81,41 @@ export const LoginOTPCard: FunctionComponent<LoginOTPCard> = ({
...
     } catch (e) {
-      setIsLoading(false);
-      alert(e);
+      if (e instanceof EnvVersionError) {
+        alert(
+          "Encountered an issue obtaining environment information. We've noted this down and are looking into it!"
+        );
```
sebastianquek, line 114: This should raise a sentry error so we can rectify it

src/components/Login/LoginOTPCard.tsx
```diff
@@ -69,11 +81,41 @@ export const LoginOTPCard: FunctionComponent<LoginOTPCard> = ({
...
+        setFeatures(versionResponse.features);
+        setProducts(versionResponse.policies);
+
+        navigation.navigate("CollectCustomerDetailsScreen");
+      } else {
+        alert(
+          "Invalid Environment Error: Make sure you scanned a valid QR code"
+        );
+        // Reset to initial login state
+        setLoginStage("SCAN");
```
sebastianquek, line 108: I suggest we pass it a `resetStage` instead, so `LoginOTPCard` doesn't need to know about the `SCAN` state

src/hooks/useLogout.tsx
dalsontws: okay, was thinking to reset them to the default values from BE, but i think an empty object would be better

src/hooks/useLogout.tsx
```diff
@@ -29,6 +30,7 @@ export const useLogout = (): LogoutHook => {
       setIsLoggingOut(true);
       await clearAuthInfo();
       setProducts([]);
+      setFeatures({} as Feature);
```
sebastianquek, line 33: Should it be the plural, `Features`?

dalsontws: yeah i think it should be haha good catch!

src/components/Login/LoginOTPCard.tsx
```diff
@@ -69,11 +75,43 @@ export const LoginOTPCard: FunctionComponent<LoginOTPCard> = ({
...
+        case "VOUCHER":
+        case "TOKEN":
+          setAuthInfo(response.sessionToken, response.ttl.getTime(), endpoint);
+          setFeatures(versionResponse.features);
+          setProducts(versionResponse.policies);
+          navigation.navigate("CollectCustomerDetailsScreen");
+          break;
+
+        // TODO: Integration with merchant flow
+        // case "MERCHANT":
```
szenius, line 96: Are we going to have separate environments for each? 

Is there a way we can generalise the two types of flows for scalability?

src/hooks/useCart/useCart.tsx
```diff
@@ -138,11 +137,6 @@ export const useCart = (
     const fetchQuota = async (): Promise<void> => {
       setCartState("FETCHING_QUOTA");
       try {
-        if (products.length === 0) {
```
szenius, line 141: Hmm I am not sure but might there be a scenario where we hit useCart and products is empty?

src/components/Login/LoginOTPCard.tsx
szenius: E.g. right now TT and CDC voucher shares the same flow. So why do we call it "Voucher"?

dalsontws: putting the same comment here: hmm yeah i think being general defo more scalable, distributions in the future (FFTH) will end up having the same flow as cdc voucher and tt token, im okay with naming the environment smth general like `COLLECTION` or `DEFAULT`?

src/hooks/useCart/useCart.tsx
dalsontws: hmm, currently the QR provides the products, and if not there's default products on the backend, unlikely we will reach useCart without products i feel!

sebastianquek: yea, given that the retrieval of products is now earlier on, there won't be the case when products is empty. Previously we wait till this part of the flow to retrieve products

src/components/Login/LoginOTPCard.tsx
```diff
@@ -69,11 +75,42 @@ export const LoginOTPCard: FunctionComponent<LoginOTPCard> = ({
...
+
+      switch (versionResponse.features.DIST_ENV) {
+        case "DEFAULT":
+          setAuthInfo(response.sessionToken, response.ttl.getTime(), endpoint);
+          setFeatures(versionResponse.features);
+          setProducts(versionResponse.policies);
+          navigation.navigate("CollectCustomerDetailsScreen");
+          break;
+
+        // TODO: Integration with merchant flow
```
szenius, line 94: Not done in this PR?

szenius: Regardless, can leave `case "MERCHANT"` uncommented actually, stacked with `default:`

src/components/Login/LoginOTPCard.tsx
```diff
@@ -69,11 +75,42 @@ export const LoginOTPCard: FunctionComponent<LoginOTPCard> = ({
...
-      navigation.navigate("CollectCustomerDetailsScreen");
+      const versionResponse = await getEnvVersion(
+        response.sessionToken,
+        endpoint
+      );
+
+      // Toggle between different environments
+      // using the DIST_ENV variable from features
+
+      switch (versionResponse.features.DIST_ENV) {
```
sebastianquek, line 86: Need to update this too

src/components/Login/LoginOTPCard.tsx
dalsontws: okay~ im leaving it here for sebs + interns to integrate, thankss!

szenius: Haha I meant the case "MERCHANT" can be uncommented actually

dalsontws: haha yeah i know! i mean im leaving it so they will do it during integration, ps for being unclear and thanks!!!🙏

PR #80, opened 2020-06-23
feat: improved keyboard avoiding view

sebastianquek:
* Add use of KeyboardAvoidingView for quota screen
* KeyboardAvoidingView acts like a normal View but when the keyboard opens, it adds a bottom padding. By doing so, the view becomes shorter. When it wraps the scroll view, elements in the scroll view will still be able to be scrolled.
* Using `behavior="padding"` only on iOS makes it work properly. On android, specifying causes issues, leaving it undefined works fine

### DIFF
--- a/src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
+++ b/src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
@@ -7,13 +7,10 @@ import React, {
 import {
   View,
   StyleSheet,
-  ScrollView,
   Keyboard,
-  KeyboardAvoidingView,
   Alert,
   Vibration,
-  BackHandler,
-  Platform
+  BackHandler
 } from "react-native";
 import { size } from "../../common/styles";
 import { Card } from "../Layout/Card";
@@ -37,6 +34,7 @@ import { FeatureToggler } from "../FeatureToggler/FeatureToggler";
 import { Banner } from "../Layout/Banner";
 import { ImportantMessageContentContext } from "../../context/importantMessage";
 import { useCheckUpdates } from "../../hooks/useCheckUpdates";
+import { KeyboardAvoidingScrollView } from "../Layout/KeyboardAvoidingScrollView";
 
 const styles = StyleSheet.create({
   content: {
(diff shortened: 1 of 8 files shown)


src/components/MerchantPayout/MerchantPayoutScreen.tsx
```diff
@@ -201,16 +201,14 @@ export const MerchantPayoutScreen: FunctionComponent<NavigationFocusInjectedProp
...
-      <ScrollView
-        contentContainerStyle={{ alignItems: "center" }}
-        scrollIndicatorInsets={{ right: 1 }}
-        keyboardShouldPersistTaps="handled"
-      >
-        <TopBackground mode={config.appMode} />
-        <KeyboardAvoidingView
-          behavior={Platform.select({ ios: "position" })}
-          keyboardVerticalOffset={Platform.select({ ios: -80 })}
+      <KeyboardAvoidingView behavior={Platform.select({ ios: "padding" })}>
```
szenius, line 204: Seems like this is always needed as a wrapper around `ScrollView`, so was wondering if we can refactor this out into a component for reuse?

dalsontws:
realised at contact number and otp screen, once keyboard appears, screen jumps up really quickly instead of smoothly moving upwards

szenius:
I don't think this is related, and not sure if it's me, but when I do `npm run start:mock` I no longer see the identifier input fields :o Works fine for stg tho

sebastianquek:
> I don't think this is related, and not sure if it's me, but when I do `npm run start:mock` I no longer see the identifier input fields :o Works fine for stg tho

Looks like the policy isn't showing up properly :O I'll look into it tomorrow!

dalsontws:
> I don't think this is related, and not sure if it's me, but when I do `npm run start:mock` I no longer see the identifier input fields :o Works fine for stg tho

eh i realised this too with `npm run start:mock`. worked for stg as well for me

src/components/MerchantPayout/MerchantPayoutScreen.tsx
sebastianquek: Good point! Previously the `KeyboardAvoidingView` was always scattered as a descendent of some `ScrollView`. Looks like it makes sense and works well when the `KeyboardAvoidingView` wraps the `ScrollView` instead.

sebastianquek:
@szenius @dalsontws I've fixed the issues, main problem was that quota and env policy was not in sync for both flows

dalsontws:
tested~ identifier input fields are there, phone number and otp screens are scrolling smoothly too! LGTM!

src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
```diff
@@ -141,40 +140,34 @@ const CollectCustomerDetailsScreen: FunctionComponent<NavigationFocusInjectedPro
 
   return (
     <>
-      <ScrollView
-        contentContainerStyle={{ alignItems: "center" }}
-        scrollIndicatorInsets={{ right: 1 }}
-        keyboardShouldPersistTaps="handled"
-      >
+      <Credits style={{ bottom: size(3) }} />
```
szenius, line 143: Ooooh can you explain why this becomes on top of the scrolling view?

src/services/envVersion/index.ts
```diff
@@ -112,7 +112,16 @@ const mockGetEnvVersion = async (
               type: "QR",
               text: "Scan"
             }
-          },
+          }
+        ]
+      },
+      {
+        category: "voucher",
+        name: "🎟️ Golden Ticket",
```
szenius, line 120: image

src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
sebastianquek: Oh because of the z-index, previously when the keyboard opens, given that the viewheight decreases, the absoluted credits will shift up and block some content

PR #85, opened 2020-06-25
feat: update display of message for otp

ddwsd:
Features
- Update alert to display messages with a different header
- Displays the warning message when the user presses the resend button that would hit the limit for OTP resends.
- Sends the user back to `LoginScanCard` if the user has hit the limit for either OTP retries or resends (and locks the user there until the 3 minute window is up).

Limitations:
- If the send/resend button that corresponds to the limit for OTP resends is pressed on the `LoginMobileNumberCard`, there would be no warnings. (I don't think this is possible given the current backend implementation)

Demo (let me know if you can't access the links!):
[OTP request error](https://gahmen.slack.com/files/U0156PY8UMC/F017JMK7N3W/otp-request-error.mp4)
[OTP request error - message passed down from mobile number card](https://gahmen.slack.com/files/U0156PY8UMC/F0182BGM396/opt-request-error-logincard.mp4)
[Account locked out](https://gahmen.slack.com/files/U0156PY8UMC/F017CF9763C/account-locked-out.mp4)
[OTP retry error](https://gahmen.slack.com/files/U0156PY8UMC/F016XNUQ4T1/otp-retry-error.mp4)

Related to [this pr](https://github.com/rationally-app/backend-api/pull/41)

### DIFF
--- a/src/components/Login/LoginContainer.tsx
+++ b/src/components/Login/LoginContainer.tsx
@@ -41,19 +41,12 @@ import { useLogout } from "../../hooks/useLogout";
 import { KeyboardAvoidingScrollView } from "../Layout/KeyboardAvoidingScrollView";
 import * as Linking from "expo-linking";
 import { DOMAIN_FORMAT } from "../../config";
-import {
-  requestOTP,
-  LoginError,
-  LoginLockedError,
-  AuthTakenError
-} from "../../services/auth";
+import { requestOTP, LoginError, LoginLockedError } from "../../services/auth";
 import {
   AlertModalContext,
   defaultConfirmationProps,
   invalidEntryAlertProps,
-  ERROR_MESSAGE,
-  disabledAccessAlertProps,
-  duplicateAlertProps
+  disabledAccessAlertProps
 } from "../../context/alert";
 
 const TIME_HELD_TO_CHANGE_APP_MODE = 5 * 1000;
@@ -143,12 +136,7 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
       lastResendWarningMessageRef.current = response.warning ?? "";
       return true;
     } catch (e) {
-      if (e instanceof AuthTakenError) {
-        showAlert({
-          ...duplicateAlertProps,
-          description: ERROR_MESSAGE.AUTH_FAILURE_TAKEN_TOKEN
-        });
-      } else if (e instanceof LoginLockedError) {
+      if (e instanceof LoginLockedError) {
         showAlert({
           ...disabledAccessAlertProps,
           description: e.message,
@@ -157,10 +145,7 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
       } else if (e instanceof LoginError) {
         showAlert({
           ...invalidEntryAlertProps,
-          description:
-            e.message === ERROR_MESSAGE.LAST_OTP_ERROR
-              ? ERROR_MESSAGE.LAST_OTP_ERROR
-              : ERROR_MESSAGE.OTP_ERROR
+          description: e.message
         });
       } else {
         alert(e);
--- a/src/components/Login/LoginContainer.tsx
+++ b/src/components/Login/LoginContainer.tsx
@@ -149,7 +149,7 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
           onOk: () => resetStage()
         });
       } else {
-        alert(e);
+        throw e;
       }
       return false;
     }
--- a/src/components/Login/LoginContainer.tsx
+++ b/src/components/Login/LoginContainer.tsx
@@ -92,7 +92,7 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
     setAllProducts
   } = useProductContext();
   const { logout } = useLogout();
-  const lastResendWarningMessageRef = useRef<string>("")
+  const lastResendWarningMessageRef = useRef<string>("");
   const resetStage = (): void => {
     setLoginStage("SCAN");
   };
--- a/src/components/Login/LoginContainer.tsx
+++ b/src/components/Login/LoginContainer.tsx
@@ -42,11 +42,7 @@ import { useLogout } from "../../hooks/useLogout";
 import { KeyboardAvoidingScrollView } from "../Layout/KeyboardAvoidingScrollView";
 import * as Linking from "expo-linking";
 import { DOMAIN_FORMAT } from "../../config";
-import {
-  requestOTP,
-  LoginError,
-  LoginLockedOutError
-} from "../../services/auth";
+import { requestOTP, LoginError, LoginLockedError } from "../../services/auth";
 
 const TIME_HELD_TO_CHANGE_APP_MODE = 5 * 1000;
 
@@ -137,7 +133,7 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
       lastResendWarningMessageRef.current = response.warning ?? "";
       return true;
     } catch (e) {
-      if (e instanceof LoginLockedOutError) {
+      if (e instanceof LoginLockedError) {
         Alert.alert(
           "Error",
           e.message,
--- a/src/components/Login/LoginContainer.tsx
+++ b/src/components/Login/LoginContainer.tsx
@@ -145,7 +145,8 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
       } else if (e instanceof LoginError) {
         showAlert({
           ...invalidEntryAlertProps,
-          description: e.message
+          description: e.message,
+          onOk: () => resetStage()
         });
       } else {
         alert(e);
--- a/src/components/Login/LoginContainer.tsx
+++ b/src/components/Login/LoginContainer.tsx
@@ -104,32 +104,34 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
     setLoginStage("SCAN");
   };
 
-  const shouldResendAlert = async (): Promise<boolean> => {
+  const getResendConfirmationIfNeeded = async (): Promise<boolean> => {
     return new Promise(resolve => {
-      Alert.alert(
-        "Resend OTP?",
-        lastResendWarningMessageRef.current,
-        [
-          {
-            text: "CANCEL",
-            onPress: () => resolve(false)
-          },
-          {
-            text: "RESEND",
-            onPress: () => resolve(true)
-          }
-        ],
-        { cancelable: false }
-      );
+      if (!lastResendWarningMessageRef.current) resolve(true);
+      else {
+        Alert.alert(
+          "Resend OTP?",
+          lastResendWarningMessageRef.current,
+          [
+            {
+              text: "CANCEL",
+              onPress: () => resolve(false)
+            },
+            {
+              text: "RESEND",
+              onPress: () => resolve(true)
+            }
+          ],
+          { cancelable: false }
+        );
+      }
     });
   };
 
-  const handleRequestOTP = async (fullNumber?: string): Promise<boolean> => {
+  const handleRequestOTP = async (
+    fullNumber = mobileNumber
+  ): Promise<boolean> => {
     try {
-      fullNumber = fullNumber ?? mobileNumber;
-      if (lastResendWarningMessageRef.current) {
-        if (!(await shouldResendAlert())) return false;
-      }
+      if (!(await getResendConfirmationIfNeeded())) return false;
       const response = await requestOTP(fullNumber, codeKey, endpointTemp);
       lastResendWarningMessageRef.current = response.warning ?? "";
       return true;
--- a/src/components/Login/LoginContainer.tsx
+++ b/src/components/Login/LoginContainer.tsx
@@ -106,8 +106,9 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
 
   const getResendConfirmationIfNeeded = async (): Promise<boolean> => {
     return new Promise(resolve => {
-      if (!lastResendWarningMessageRef.current) resolve(true);
-      else {
+      if (!lastResendWarningMessageRef.current) {
+        resolve(true);
+      } else {
         Alert.alert(
           "Resend OTP?",
           lastResendWarningMessageRef.current,
@@ -136,16 +137,27 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
       lastResendWarningMessageRef.current = response.warning ?? "";
       return true;
     } catch (e) {
-      if (e instanceof LoginError) {
+      if (e instanceof LoginLockedOutError) {
         Alert.alert(
           "Error",
           e.message,
           [
             {
               text: "OK",
-              onPress: () => {
-                if (e instanceof LoginLockedOutError) resetStage();
-              }
+              onPress: () => resetStage()
+            }
+          ],
+          {
+            cancelable: false
+          }
+        );
+      } else if (e instanceof LoginError) {
+        Alert.alert(
+          "Error",
+          e.message,
+          [
+            {
+              text: "OK"
             }
           ],
           {
--- a/src/components/Login/LoginContainer.tsx
+++ b/src/components/Login/LoginContainer.tsx
@@ -293,7 +293,7 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
               setMobileNumber={setMobileNumber}
               codeKey={codeKey}
               endpoint={endpointTemp}
-              lastResendWarningMessage={lastResendWarningMessageRef}
+              lastResendWarningMessageRef={lastResendWarningMessageRef}
             />
           )}
           {loginStage === "OTP" && (
@@ -302,7 +302,7 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
               mobileNumber={mobileNumber}
               codeKey={codeKey}
               endpoint={endpointTemp}
-              lastResendWarningMessage={lastResendWarningMessageRef}
+              lastResendWarningMessageRef={lastResendWarningMessageRef}
             />
           )}
           <FeatureToggler feature="HELP_MODAL">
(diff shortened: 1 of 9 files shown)


src/components/Login/LoginContainer.tsx
```diff
@@ -89,6 +89,7 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
   const resetStage = (): void => {
     setLoginStage("SCAN");
   };
+  const [lastResendWarningMessage, setLastResendWarningMessage] = useState("");
```
dalsontws, line 92: can you elaborate on why we need the use of state hook to set the resend error message?

ddwsd: let me try using something else!

src/components/Login/LoginOTPCard.tsx
```diff
@@ -95,6 +107,22 @@ export const LoginOTPCard: FunctionComponent<LoginOTPCard> = ({
         alert(
           "Encountered an issue obtaining environment information. We've noted this down and are looking into it!"
         );
+      } else if (e instanceof LoginError) {
+        Alert.alert(
```
ilhaoni, line 111: Should Sentry capture the exception here too?

ddwsd: hmm that would mean it captures an exception every time someone keys in the wrong otp. which may be very often  haha. it didnt capture previously

src/components/Login/LoginOTPCard.tsx
```diff
@@ -35,20 +40,23 @@ interface LoginOTPCard {
   mobileNumber: string;
   codeKey: string;
   endpoint: string;
+  lastResendWarningMessageRef: MutableRefObject<string>;
```
dalsontws, line 43: slightly confused with the use of a `MutableRefObject` here for the warning message, can you elaborate a bit on this?

ddwsd: the `MutableRefObject` remembers the data stored in the warning message through rerenders. does it make the code a lot less readable? i can change it to be a variable outside the `LoginContainer` component if so i think

dalsontws: it seems slightly complex for the warning message, whats the data stored in the warning message?

ddwsd: it's just a string 😅

ilhaoni: hmm, this is to be able to pass the message into the individual card ah?

src/components/Login/LoginOTPCard.tsx
ilhaoni: I see I see, okok, let's leave it then haha

src/components/Login/LoginOTPCard.tsx
ddwsd: yeahh

ilhaoni: hmm, is there a reason we have to keep the last message state actually? is it due to some checks haha

ilhaoni: ahh, is it for the alert before resend?

ddwsd: yup this is for the case where the user moves on to the `LoginOTPCard` and the message was triggered by an entry in the `LoginMobileNumberCard`

dalsontws: would it be better for it to be stored just as variable outside `LoginContainer`? and also, the warning messages will always be for the case of requesting too many OTPs?

ddwsd: the warning message will warn the user just before he requests the last OTP before he/she times out. the message is as follows
> After 1 more OTP request, you will have to wait 3 minutes before trying again.

> would it be better for it to be stored just as variable outside LoginContainer? 
sure, i can try that

ddwsd: hmm i tried it and it didn't work. i was following [this article](https://www.emgoto.com/storing-values-with-useref/) when i suggested that. i guess it's not that easy to change parent state with child components

ilhaoni: ahhh, i see i see. Hmm, maybe let's leave this for now? We can create a tech chore to remind ourselves to look into this to see if there's a better way haha @sebastianquek what do you think ahah

dalsontws: also @ddwsd i think for this we can add the new AlertModal we are using!!

src/components/Login/LoginContainer.tsx
```diff
@@ -40,11 +41,14 @@ import { useLogout } from "../../hooks/useLogout";
 import { KeyboardAvoidingScrollView } from "../Layout/KeyboardAvoidingScrollView";
 import * as Linking from "expo-linking";
 import { DOMAIN_FORMAT } from "../../config";
+import { requestOTP, LoginError } from "../../services/auth";
 
 const TIME_HELD_TO_CHANGE_APP_MODE = 5 * 1000;
 
 const ALLOW_MODE_CHANGE = false;
 
+let lastResendWarningMessage = "";
```
sebastianquek, line 50: Why do we still need this variable?

src/components/Login/LoginOTPCard.tsx
```diff
@@ -141,7 +177,9 @@ export const LoginOTPCard: FunctionComponent<LoginOTPCard> = ({
           ) : (
             <SecondaryButton
               text="Resend"
-              onPress={resendOTP}
+              onPress={
+                lastResendWarningMessage === "" ? resendOTP : alertBeforeResend
+              }
```
sebastianquek, line 182: Why not let `LoginContainer` handle whether to show the resend warning message? Recall we want to show this on pressing submit after entering phone number too right?

src/services/helpers.ts
```diff
@@ -15,7 +15,6 @@ export async function fetchWithValidator<T, O, I>(
   init?: RequestInit
 ): Promise<T> {
   const response = await fetch(requestInfo, init);
-
```
sebastianquek, line 18: Do discard this ya

src/components/Login/LoginOTPCard.tsx
```diff
@@ -108,14 +131,27 @@ export const LoginOTPCard: FunctionComponent<LoginOTPCard> = ({
...
-      setIsResending(false);
-      alert(e.message || e);
-    }
+    const isRequestSuccessful = await handleRequestOTP();
+    setIsResending(false);
+    if (isRequestSuccessful) setResendDisabledTime(RESEND_OTP_TIME_LIMIT);
+  };
+
+  const alertBeforeResend = (): void => {
+    console.log(lastResendWarningMessage);
```
sebastianquek, line 140: Do remove this too.

src/components/Login/LoginOTPCard.tsx
ddwsd: hmm the alert should not show after entering phone number for now, but yeah ill move it up in case this is supported in the future

src/components/Login/LoginContainer.tsx
ddwsd: it's to store the warning message sent from the backend, because it won't be shown immediately, but only after the user presses resend again

src/services/auth/auth.test.tsx
```diff
@@ -39,6 +41,16 @@ describe("auth", () => {
       ]);
     });
 
+    it("should return an object with warning and status keys if the response contains these properties", async () => {
```
sebastianquek, line 44: Possible to include a test to express what the difference in these 2 properties are?

src/components/Login/LoginContainer.tsx
```diff
@@ -91,6 +99,59 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
...
+            text: "RESEND",
+            onPress: () => resolve(true)
+          }
+        ],
+        { cancelable: false }
+      );
+    });
+  };
+
+  const handleRequestOTP = async (fullNumber?: string): Promise<boolean> => {
```
sebastianquek, line 122: What if we default the `fullNumber` to `mobileNumber` rather than relying on `fullNumber ?? mobileNumber`?

src/components/Login/LoginContainer.tsx
```diff
@@ -91,6 +99,59 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
...
+    try {
+      fullNumber = fullNumber ?? mobileNumber;
+      if (lastResendWarningMessage) {
```
sebastianquek: Do we need this condition? Can we move this into the `shouldResendAlert`?

I'm thinking the concept of the `shouldResendAlert` is closer to `getResendConfirmationIfNeeded`, so the consumer just calls this and if it's true, proceed.

src/components/Login/LoginContainer.tsx
```diff
@@ -91,6 +99,59 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
...
+    } catch (e) {
+      if (e instanceof LoginError) {
+        Alert.alert(
+          "Error",
+          e.message,
+          [
+            {
+              text: "OK",
+              onPress: () => {
+                if (e instanceof LoginLockedOutError) resetStage();
```
sebastianquek, line 140: Hmm, wonder if it's better to move this out so we treat `LoginError` and `LoginLockedOutError` as siblings. This provides flexibility when we want to add other types of LoginErrors that may not have the same format as this alert where there's only 1 OK button

src/components/Login/LoginMobileNumberCard.tsx
```diff
@@ -54,16 +51,11 @@ export const LoginMobileNumberCard: FunctionComponent<LoginMobileNumberCard> = (
...
-      setIsLoading(false);
-      setMobileNumber(fullNumber);
-      setLoginStage("OTP");
-    } catch (e) {
-      setIsLoading(false);
-      alert(e);
-    }
+    const fullNumber = createFullNumber(countryCode, mobileNumberValue);
+    const isRequestSuccessful = await handleRequestOTP(fullNumber);
+    setMobileNumber(fullNumber);
```
sebastianquek, line 56: Do we still want to set mobile number if the request of OTP fails? What are the advantages of using this true/false approach as compared to the try/catch approach?

src/components/Login/LoginOTPCard.tsx
```diff
@@ -95,6 +102,22 @@ export const LoginOTPCard: FunctionComponent<LoginOTPCard> = ({
         alert(
           "Encountered an issue obtaining environment information. We've noted this down and are looking into it!"
         );
+      } else if (e instanceof LoginError) {
```
sebastianquek, line 105: Clarification: Are the errors being thrown from `validateOTP`?

sebastianquek (changes requested):

src/components/Login/LoginContainer.tsx
ddwsd: i think if there are other types of LoginErrors we can just capture them separately?

src/components/Login/LoginOTPCard.tsx
ddwsd: yup!

src/components/Login/LoginContainer.tsx
ddwsd: i think thats better

src/components/Login/LoginMobileNumberCard.tsx
ddwsd: i was following the previous code when i did this, but yeah i don't think we should if the request fails.
i think this approach moves all the reusable error handling logic to the `LoginContainer`

src/components/Login/LoginContainer.tsx
ddwsd: i see. that sounds better, thanks!!

src/services/auth/auth.test.tsx
ddwsd: i see. didn't know that this is a use for tests

src/services/auth/index.tsx
```diff
@@ -11,15 +10,22 @@ export class LoginError extends Error {
   }
 }
 
+export class LoginLockedOutError extends LoginError {
```
ilhaoni, line 13: Small nit, this is a bit of a mouthful. What if we reduce it to `LockedOutError` ah? Any issues? 😆

src/components/Login/LoginContainer.tsx
```diff
@@ -99,32 +99,34 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
...
-          },
-          {
-            text: "RESEND",
-            onPress: () => resolve(true)
-          }
-        ],
-        { cancelable: false }
-      );
+      if (!lastResendWarningMessageRef.current) resolve(true);
+      else {
```
sebastianquek, line 105: Minor nit: can we use the braces here since we're having the else clause already.

src/services/auth/index.tsx
sebastianquek: Sure am good with this

src/components/Login/LoginContainer.tsx
sebastianquek: Yep, why is `LoginLockedOutError` a special case then? Can it be treated in a block on its own?

src/services/auth/index.tsx
ddwsd: it's less descriptive i guess, like i might not know right away that it is related to login. what about `LoginLockedError`?😅

ilhaoni: Hmm, that's true too. I'm good with `LoginLockedError` haha. Seb? ^^

src/components/Login/LoginContainer.tsx
ddwsd: hmm...for the current alerts i think treating it in its own block is a bit verbose. but once we change to using the alert modals it won't be that much of an issue. will change it here then!

src/services/auth/auth.test.tsx
sebastianquek: What I meant was to provide more details on what these properties mean and how they interact with each other. You've added a test showcasing that we only need `status` and not `warning`, but looking at the types, what if there's only `warning` and what if the response is empty?

These can either be defined in the tests or as comments in the defined type. Both provide a means to let future devs know what kind of results are expected. The tests explicitly showcase some possible values, e.g. is status a status code? is it a human readable string? The types on the other hand showcase which must exist in the response (i.e required vs optional).

src/services/auth/index.tsx
sebastianquek: sure

src/components/Login/LoginContainer.tsx
sebastianquek: Okay!

src/services/auth/auth.test.tsx
ddwsd: i see. looking at the backend code, status should always be returned. will update the tests and types accordingly

src/components/Login/LoginContainer.tsx
ddwsd: will update the alerts to use the alert modals!

ddwsd:
currently this PR shows a generic "Wrong entry" title for all `LoginError`s other than the `LoginLockedError`. will update the titles in a future PR (that would probably have types for the [different errors thrown in the backend](https://github.com/rationally-app/backend-api/pull/109))

src/components/Login/LoginContainer.tsx
```diff
@@ -91,11 +99,61 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
...
+        showAlert({
+          ...disabledAccessAlertProps,
+          description: e.message,
+          onOk: () => resetStage()
+        });
+      } else if (e instanceof LoginError) {
+        showAlert({
+          ...invalidEntryAlertProps,
+          description: e.message
+        });
```
sebastianquek, line 149: What is the default CTA for this? Do we want to reset the stage?

src/components/Login/LoginContainer.tsx
```diff
@@ -91,11 +99,61 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
...
+      } else {
+        alert(e);
```
sebastianquek, line 151: If it's an unknown error, do you think it'll be useful to show the ErrorBoundary or just show a generic warning alert?

src/components/Login/LoginContainer.tsx
ddwsd: hmm i think in most cases, a `LoginError` is thrown here when the user is using an already used QR code, or a QR code that is valid but not inside our database. it does make sense to reset the stage in these cases. i can set that to be the default CTA here, and in a future PR where we have finer handling of login errors we can have more appropriate CTAs for each cases

src/components/Login/LoginContainer.tsx
ddwsd: yeah, i think ErrorBoundary is more useful. will do that in another pr!

sebastianquek (approved):
LGTM!

ilhaoni (approved):
LGTM~~ Thanks for the great work!!

ilhaoni:
@dalsontws do you wanna do a last review? ^^

src/components/Login/LoginOTPCard.tsx
```diff
@@ -102,6 +116,17 @@ export const LoginOTPCard: FunctionComponent<LoginOTPCard> = ({
...
+          description: e.message,
+          onOk: () => resetStage()
+        });
+      } else if (e instanceof LoginError) {
+        showAlert({
+          ...invalidEntryAlertProps,
+          description: e.message
+        });
       } else {
         alert(e);
```
dalsontws, line 131: for this, can throw an error? so that if its unhandled, it will reach the error boundary. 
can also let have sentry capture it too

ddwsd: yeah, i think it should be in another pr where we standardise the login error dialogs tho

ddwsd: ok nvm i'll do it here!

PR #86, opened 2020-06-25
feat: identifier validation based on regex

szenius:
This PR:
* Adds identifier validation based on `validationRegex` passed in `PolicyIdentifier`
* Refactors identifier validations into a single util function
* Adds unit tests for the new validations

Things on the side:
* Made `InputWithLabel.editable` optional -- this was added to support disabled input, but the original uses of this component didn't pass in an `editable` param

Can be tested with [PR49](https://github.com/rationally-app/backend-api/pull/49).

### DIFF
--- a/src/utils/validateIdentifierInputs.ts
+++ b/src/utils/validateIdentifierInputs.ts
@@ -8,10 +8,20 @@ const isMatchRegex = (text: string, regex?: string): boolean => {
   return new RegExp(regex).test(text);
 };
 
+const isUniqueList = (list: string[]): boolean =>
+  new Set(list).size === list.length;
+
 export const validateIdentifierInputs = (
   identifierInputs: IdentifierInput[]
 ): boolean => {
   for (const { value, validationRegex, textInputType } of identifierInputs) {
+    if (!value) {
+      throw new Error(
+        `Please enter ${
+          identifierInputs.length === 1 ? "" : "unique "
+        }details to checkout`
+      );
+    }
     if (textInputType === "NUMBER" && isNaN(Number(value))) {
       throw new Error("Invalid details");
     }
@@ -22,5 +32,18 @@ export const validateIdentifierInputs = (
       throw new Error("Invalid contact number");
     }
   }
+
+  if (
+    !isUniqueList(
+      identifierInputs.map(identifierInput => identifierInput.value)
+    )
+  ) {
+    throw new Error(
+      `Please enter ${
+        identifierInputs.length === 1 ? "" : "unique "
+      }details to checkout`
+    );
+  }
+
   return true;
 };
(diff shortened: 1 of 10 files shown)


src/utils/validatePhoneNumbers.tsx
```diff
@@ -32,12 +32,3 @@ export const countryCodeValidator = (code: string): boolean => {
   );
   return code[0] === "+" && countryCodesList.includes(code.substring(1));
 };
-
-export const validatePhoneNumbers = (phoneNumbers: string[]): boolean => {
```
sebastianquek, line 36: Hmmm is this being used by the OTP flow?

szenius: Not this function. OTP flow uses other functions in this file. I created this for the checkout previously~

src/utils/validateIdentifierInputs.ts
```diff
@@ -0,0 +1,26 @@
...
+    if (textInputType === "NUMBER" && isNaN(Number(value))) {
+      throw new Error("Invalid details");
+    }
+    if (!isMatchRegex(value, validationRegex)) {
+      throw new Error("Invalid details");
+    }
+    if (textInputType === "PHONE_NUMBER" && !fullPhoneNumberValidator(value)) {
+      throw new Error("Invalid contact number");
+    }
+  }
```
sebastianquek, line 24: Do you think there's a need to validate blank `value`s too?

szenius: It's validated in a block before this in `checkoutCart`. Didn't add it here because there are other conditions sharing the same error message. Not sure which is better 🤔

sebastianquek: Oh i see, there's the check for `numUnverifiedTransactions`. Hmm, thinking that since this function can be used anywhere, should add the safeguards too, what do you think?

szenius: Makes sense, will move that check here!

szenius: Refactored :D

src/utils/validateIdentifierInputs.ts
```diff
@@ -0,0 +1,49 @@
...
+  for (const { value, validationRegex, textInputType } of identifierInputs) {
+    if (!value) {
+      throw new Error(
+        `Please enter ${
+          identifierInputs.length === 1 ? "" : "unique "
+        }details to checkout`
+      );
+    }
+    if (textInputType === "NUMBER" && isNaN(Number(value))) {
+      throw new Error("Invalid details");
```
ilhaoni, line 26: small nit, i was wondering whether it would be clearer if it's `Invalid field value`, but its ok HAHA, merged liao 😂

szenius: Oops sorryy, I followed the "details" copy that is currently in the app (when you don't fill in, it says "Please enter details to check out") But will check with Eida again

ilhaoni: HAHA no lah, dont need to say sorry haha. Okk~~ \^^/

szenius: Eida will create separate card for all the different copy changes for prioritisation~

PR #89, opened 2020-06-29
feat: no negative quota

ddwsd:
[Trello](https://trello.com/c/1p4FC27Q/192-ensure-app-doesnt-show-a-cart-with-negative-quantity)
Features:
- Check for negative quota received and display the limit reached screen if there are
- Sends a Sentry error ([without stacktrace](https://docs.sentry.io/error-reporting/capturing/?platform=javascript#capturing-errors--exceptions))
- Added tests for this case

### DIFF
--- a/src/hooks/useCart/useCart.test.tsx
+++ b/src/hooks/useCart/useCart.test.tsx
@@ -338,8 +338,8 @@ describe("useCart", () => {
           category: "toilet-paper",
           identifierInputs: [],
           lastTransactionTime: transactionTime,
-          maxQuantity: -1,
-          quantity: -1
+          maxQuantity: 0,
+          quantity: 0
         },
         {
           category: "chocolate",
--- a/src/hooks/useCart/useCart.test.tsx
+++ b/src/hooks/useCart/useCart.test.tsx
@@ -184,6 +184,24 @@ const mockQuotaResSingleIdNoQuota: Quota = {
     }
   ]
 };
+
+const mockQuotaResSingleIdNegativeQuota: Quota = {
+  remainingQuota: [
+    {
+      category: "toilet-paper",
+      identifierInputs: [],
+      quantity: -1,
+      transactionTime
+    },
+    {
+      category: "chocolate",
+      identifierInputs: [],
+      quantity: 15,
+      transactionTime
+    }
+  ]
+};
+
 const mockQuotaResMultipleIds: Quota = {
   remainingQuota: [
     {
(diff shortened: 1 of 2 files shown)


src/hooks/useCart/useCart.test.tsx
```diff
@@ -301,7 +319,36 @@ describe("useCart", () => {
...
+
+      await waitForNextUpdate();
+      expect(result.current.cartState).toBe("NO_QUOTA");
+      expect(result.current.cart).toStrictEqual([
+        {
+          category: "toilet-paper",
+          identifierInputs: [],
+          lastTransactionTime: transactionTime,
+          maxQuantity: -1,
+          quantity: -1
```
sebastianquek, line 341: Although it's not used once the app reaches this state, the cart should be more resilient by not allowing negative values. So if for whatever reason the cart is seen, the user will never see negative values.

src/hooks/useCart/useCart.tsx
```diff
@@ -144,7 +148,10 @@ export const useCart = (
           setFeatures(response.features);
         }
         const quotaResponse = await getQuota(ids, authKey, endpoint);
-        if (hasNoQuota(quotaResponse)) {
+        if (hasNegativeQuota(quotaResponse)) {
+          Sentry.captureException("Negative Quota Received");
+          setCartState("NO_QUOTA");
+        } else if (hasNoQuota(quotaResponse)) {
```
sebastianquek, line 154: The negative quota case is really an outlier and practically never happens unless we made a mistake on the backend. Furthermore, the concept of `hasNegativeQuota` is strange in a practical sense, implies that the user owes some items. We should treat the idea of non positive quota as having no quota.

I suggest we move the logic into `hasNoQuota`, so future developers don't need to worry about the concept of "negative quota" because in the first place it should not happen. In a sense, we relegate the notion of "negative quota" to be within the concept of "no quota"

So in the `every` iterator, we can check if it's negative and just log a sentry issue

ddwsd: these are good points, thanks!

ddwsd:
I thought for a while about whether I should set the quantity to 0 when a negative quantity is received, or leave it unchanged and just display 0 to the user. I came to the conclusion that since negative quota should be regarded as no quota, I should set it to 0 so possible future features like sums won't be affected. Let me know what you think, thanks!

src/hooks/useCart/useCart.tsx
```diff
@@ -107,8 +108,18 @@ const mergeWithCart = (
...
+  for (const item of quota.remainingQuota) {
+    // The negative quantity case should never occur but if it does, we send an error to Sentry and set state to NO_QUOTA
+    if (item.quantity < 0) {
+      Sentry.captureException("Negative Quota Received");
+      item.quantity = 0;
+      hasNegative = true;
+    } else result = result && item.quantity === 0;
+  }
+  return result || hasNegative;
+};
```
sebastianquek, line 122: Couple of guiding questions to improve this:

1. Given `hasNoQuota` returns a boolean, will a consumer know that it actually has side effects (such as changing item quantity to 0)? Why should this function have side effects? Could there be a better approach?
2. How might we make use of existing TS constructs to further simplify the code? Can we make use of existing TS array functions (e.g. `every`) to help achieve the same outcome?

Let me know if you have questions!

src/hooks/useCart/useCart.tsx
```diff
@@ -96,10 +97,10 @@ const mergeWithCart = (
         return {
           category,
           quantity: Math.min(
-            maxQuantity,
+            maxQuantity < 0 ? 0 : maxQuantity,
             existingItem?.quantity || defaultQuantity
           ),
-          maxQuantity,
+          maxQuantity: maxQuantity < 0 ? 0 : maxQuantity,
```
sebastianquek, line 103: Can make use of `Math.max` for these

src/hooks/useCart/useCart.tsx
```diff
@@ -138,11 +143,12 @@ export const useCart = (
       setCartState("FETCHING_QUOTA");
       try {
         const quotaResponse = await getQuota(ids, authKey, endpoint);
-        if (hasNoQuota(quotaResponse)) {
+        if (hasInvalidQuota(quotaResponse)) {
+          Sentry.captureException("Negative Quota Received");
```
sebastianquek, line 147: Come to think of it, it'll be good to specify the quota when capturing the exception too

src/hooks/useCart/useCart.tsx
```diff
@@ -138,11 +143,16 @@ export const useCart = (
...
+              quotaResponse.remainingQuota
+            )}`
+          );
           setCartState("NO_QUOTA");
-        } else {
-          setCartState("DEFAULT");
-        }
+        } else if (hasNoQuota(quotaResponse)) {
+          setCartState("NO_QUOTA");
+        } else setCartState("DEFAULT");
```
sebastianquek, line 155: Minor nitpick, it's more readable if we follow the convention of having all conditions in curly brackets
```suggestion
        } else {
          setCartState("DEFAULT");
        }
```

PR #90, opened 2020-06-30
fix: add fetching of envVersion in useCart

dalsontws:


dalsontws:
closing PR since this fix is no longer required

PR #93, opened 2020-06-30
feat: fetch version and features during redirect

dalsontws:
[trello](https://trello.com/c/3i74wCKH)

### DIFF
--- a/src/components/Login/LoginContainer.tsx
+++ b/src/components/Login/LoginContainer.tsx
@@ -36,6 +36,7 @@ import { FeatureToggler } from "../FeatureToggler/FeatureToggler";
 import { ImportantMessageContentContext } from "../../context/importantMessage";
 import { Banner } from "../Layout/Banner";
 import { getEnvVersion, EnvVersionError } from "../../services/envVersion";
+import { useProductContext } from "../../context/products";
 
 const TIME_HELD_TO_CHANGE_APP_MODE = 5 * 1000;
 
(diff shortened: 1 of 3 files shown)


src/components/Login/LoginOTPCard.tsx
```diff
@@ -88,11 +88,6 @@ export const LoginOTPCard: FunctionComponent<LoginOTPCard> = ({
           setAuthInfo(response.sessionToken, response.ttl.getTime(), endpoint);
           setFeatures(versionResponse.features);
           setProducts(versionResponse.policies);
-        case "MERCHANT":
```
sebastianquek, line 91: Hmm thinking if we want to just check for the existence of FLOW_TYPE rather than switch thru the different envs here, since actually at this point this component doesnt care, it just wants to know whether to save the info or not.

src/components/Login/LoginContainer.tsx
```diff
@@ -87,11 +90,42 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
...
+          break;
+        case "MERCHANT":
+          navigation.navigate("MerchantPayoutScreen");
+          break;
+        default:
+          alert(
+            "Invalid Environment Error: Make sure you scanned a valid QR code"
+          );
+          // Reset to initial login state
+          () => setLoginStage("SCAN");
```
sebastianquek, line 125: I think should just call the function?

src/components/Login/LoginOTPCard.tsx
dalsontws: agree with this, since envs are check in login container, any possibility that the FLOW_TYPE is some invalid string?

sebastianquek: Technically it will be invalid if we change the types of strings accepted (e.g. COLLECTION vs MERCHANT). But then again, this switch just checks whether to save info or not. So regardless, once FLOW_TYPE is defined, I think it's safe to say that the info should be saved

PR #94, opened 2020-07-01
feat: add validation for expiry, logout when error thrown 

dalsontws:


src/components/Login/LoginContainer.tsx
```diff
@@ -107,15 +115,25 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
...
           );
+          handleLogout();
         }
       } finally {
         setIsLoading(false);
       }
     };
-    if (token && endpoint && !features) {
+    const diffInSeconds = differenceInSeconds(Number(expiry), Date.now());
+    if (token && endpoint && diffInSeconds > 0 && !features) {
```
sebastianquek, line 125: Can use `isAfter` or `isBefore` instead since we're just checking for `> 0`.

Also, shall we call logout if expired already? So user won't be able to continue further if token has expired.

dalsontws: yeap ok added that in! actually would the use of effect hook to validate expiry be better here? or this works fine haha

sebastianquek: I just discovered, we're actually calling `validateTokenExpiry` in `Content`, so actually on launching the app, we already check expiry (by checking local expiry).

So we don't need this expiry check at all. Can just handle the logout on error

src/components/Login/LoginContainer.tsx
```diff
@@ -115,7 +123,15 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
...
     }
-  }, [endpoint, token, setFeatures, setProducts, features]);
+  }, [
+    endpoint,
+    token,
+    setFeatures,
+    setProducts,
+    features,
+    handleLogout,
+    expiry
```
sebastianquek, line 133: Don't need this dependency i think

dalsontws: oops thanks! also removed navigation attribute from loginOTPcard

PR #96, opened 2020-07-02
feat: update merchant code validation

sebastianquek:
8 max limit alphanumeric (means if they enter 6 characters can go through too)

### DIFF
--- a/src/utils/validateMerchantCode.test.tsx
+++ b/src/utils/validateMerchantCode.test.tsx
@@ -2,16 +2,20 @@ import { validateMerchantCode } from "./validateMerchantCode";
 
 describe("validate merchant code", () => {
   it("should return true for valid merchant code", () => {
-    expect.assertions(5);
-    expect(validateMerchantCode("100000001")).toBe(true);
-    expect(validateMerchantCode("202006171")).toBe(true);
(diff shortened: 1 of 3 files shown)


src/utils/validateMerchantCode.tsx
```diff
@@ -1,4 +1,4 @@
-export const merchantCodeRegex = /^[a-zA-Z0-9 _-]+$/; // Alphanumeric, space, underscore, dash
+export const merchantCodeRegex = /^[a-zA-Z0-9]{1,8}$/;
```
szenius, line 1: Why can't we use policy validationRegex for this?

sebastianquek: Yea was thinking we stick to this first, once the backend PR merges, we can change it!

src/utils/validateMerchantCode.test.tsx
```diff
@@ -22,5 +28,8 @@ describe("validate merchant code", () => {
     expect(() => validateMerchantCode("å∫ç∂´ƒ©˙")).toThrow(
       "Invalid merchant code"
     );
+    expect(() => validateMerchantCode("ABCD12345678")).toThrow(
```
dalsontws, line 31: just my thoughts, maybe can cater for an edge case like 9 characters?

PR #101, opened 2020-07-13
feat: request timeout set at 60s

dhemutton:
This PR contains the timeout for fetch but only applied to the quota service in case it breaks the other services. Referenced this: https://github.com/github/fetch/issues/175


sebastianquek:
Please remove your changes to package-lock

src/services/helpers.ts
```diff
@@ -16,6 +16,36 @@ export async function fetchWithValidator<T, O, I>(
...
+}
+
+async function timeout<Response>(ms: number, promise: any): Promise<Response> {
+  return new Promise<Response>(async (resolve, reject) => {
+    setTimeout(() => {
+      reject(new Error("Request timed out. Please try again."));
+    }, ms);
+    resolve(await promise);
+  });
+}
```
sebastianquek, line 41: Given that this is a generic function and not confined to only `fetch`ing, we can make a couple of improvements:
1. Use generics for the type rather than using `Response`, the param of promise also doesn't need to be `any`
2. Using a `TimeoutError` will be useful as we propagate the error up the stack. We can perhaps handle it differently in future. e.g. when timeout error occurs, we show a different message
3. Given that this function may be used in different contexts, we can further make it customisable by leaving the message as an optional param. Also, I suggest we change it to be more actionable: "Request timed out. Please check your internet connection and try again later."

src/services/helpers.ts
```diff
@@ -16,6 +16,36 @@ export async function fetchWithValidator<T, O, I>(
...
+
+export async function fetchWithValidatorWithTimeOut<T, O, I>(
+  validator: Type<T, O, I>,
+  requestInfo: RequestInfo,
+  init?: RequestInit
+): Promise<T> {
+  const response = await timeout<Response>(60000, fetch(requestInfo, init));
```
sebastianquek, line 48: Can extract the timeout duration to an env variable!

justussoh: 👍

src/services/helpers.ts
```diff
@@ -9,13 +9,58 @@ export class ValidationError extends Error {
...
+): Promise<T> {
+  return new Promise<T>(async (resolve, reject) => {
+    setTimeout(() => {
+      reject(new TimeoutError(errorMessage || "Request timed out"));
+    }, ms);
+    resolve(await promise);
+  });
+}
+
+export async function fetchWithValidatorWithTimeOut<T, O, I>(
```
sebastianquek, line 54: Come to think of it, we don't need an additional function, we can add a `timeout?: boolean | number` param to `fetchWithValidator`, that defaults to `false`. So when we want a fetch to timeout, we can either add `true` (which defaults to `API_CALL_TIMEOUT` or the number of milliseconds.

package-lock.json
```diff
@@ -16247,4 +16247,4 @@
       }
     }
   }
-}
```
sebastianquek, line 16250: Discard this change too 😆

src/services/helpers.ts
```diff
@@ -1,5 +1,5 @@
 import { fold } from "fp-ts/lib/Either";
-import { Type } from "io-ts";
+import { Type, number } from "io-ts";
```
sebastianquek, line 2: Don't need this `number` import

src/services/helpers.ts
```diff
@@ -9,14 +9,47 @@ export class ValidationError extends Error {
...
+    resolve(await promise);
+  });
+}
+
 export async function fetchWithValidator<T, O, I>(
   validator: Type<T, O, I>,
   requestInfo: RequestInfo,
-  init?: RequestInit
+  init?: RequestInit,
+  timeout: boolean | number = false
```
sebastianquek, line 36:
```suggestion
  timeout?: boolean | number = false
```

src/services/helpers.ts
```diff
@@ -9,14 +9,47 @@ export class ValidationError extends Error {
...
 ): Promise<T> {
-  const response = await fetch(requestInfo, init);
+  let response;
+  if (timeout) {
+    const timeoutVal =
+      typeof timeout === "number" ? timeout : process.env.API_CALL_TIMEOUT;
```
sebastianquek, line 41: Make sure to default `process.env.API_CALL_TIMEOUT` too, since it may not be defined

src/services/quota/index.tsx
```diff
@@ -1,6 +1,6 @@
 import { IS_MOCK } from "../../config";
 import { Transaction, Quota, PostTransactionResult } from "../../types";
-import { fetchWithValidator, ValidationError } from "../helpers";
+import { fetchWithValidatorWithTimeOut, ValidationError } from "../helpers";
```
sebastianquek, line 3: Remember to update this

src/services/helpers.ts
dhemutton: oo seb i got an error "Parameter cannot have question mark and initializer.ts(1015)" when i did that because i think it's initialised as false

src/services/helpers.ts
dhemutton: i think we don't have to worry cos timeout is default false without the "?" 
Screenshot 2020-07-16 at 2 09 37 PM

src/services/helpers.ts
```diff
@@ -9,14 +10,51 @@ export class ValidationError extends Error {
...
+      reject(new TimeoutError(errorMessage || "Request timed out"));
+    }, ms);
+    resolve(await promise);
+  });
+}
+
+// If timeout
+// true => timeout duration for fetch request will be set as env var.
+// false => usual fetch request is done.
+// TimeoutMS (eg. 10) => timeout duration for fetch request will be set as given val in milliseconds.
```
sebastianquek, line 36: Btw you can do something like this instead: 
image

Then it will appear nicely in the intellisense:
image

src/services/helpers.ts
sebastianquek: Oh right, there's no path that will get a possibly undefined process env

src/types.ts
```diff
@@ -154,3 +154,4 @@ export type Voucher = {
   serial: string;
   denomination: number;
 };
+export type TimeoutMS = number;
```
sebastianquek, line 157: Hmm i think we can move this back into helper.ts since it's not shared elsewhere

src/services/quota/index.tsx
```diff
@@ -127,6 +127,7 @@ export const liveGetQuota = async (
     if (e instanceof ValidationError) {
       Sentry.captureException(e);
     }
+
```
sebastianquek, line 130: Do discard this haha, so when we git blame, it won't show up incorrectly

src/services/helpers.ts
sebastianquek: Ok, whoops

src/services/helpers.ts
sebastianquek: @dhemutton don't forget this change too 😆

sebastianquek:
Given the difficulty in setting a good request timeout plus the timeout doesn't actually call a cancellation, we should revisit this in future again when we encounter long requests.

PR #102, opened 2020-07-14
feat: updated cache timeout to 10s

dhemutton:
This PR contains the increase in time out duration to 4seconds for automatic updates over-the-air (OTA): https://docs.expo.io/guides/configuring-ota-updates/#automatic-updates

Screenshot 2020-07-14 at 11 54 52 AM

### DIFF
--- a/app.json
+++ b/app.json
@@ -15,7 +15,7 @@
       "backgroundColor": "#F1FAFA"
     },
     "updates": {
-      "fallbackToCacheTimeout": 0
+      "fallbackToCacheTimeout": 10000
     },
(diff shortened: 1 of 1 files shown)


app.json
```diff
@@ -15,9 +19,11 @@
       "backgroundColor": "#F1FAFA"
     },
     "updates": {
-      "fallbackToCacheTimeout": 0
+      "fallbackToCacheTimeout": 4
```
sebastianquek, line 22: I think this is in milliseconds? Also let's do 10 seconds, and please commit only this line :p

PR #103, opened 2020-07-14
feat: using new feature flag to toggle id scanning type

dalsontws:
[notion](https://www.notion.so/Feature-Identification-scanning-Using-feature-flag-to-toggle-id-scanning-type-448e2acc9bb644dd9dbbaf38923d8960)

Proposed Solution:
Adding a new feature flag for scanner, toggling the type and validation
```
id: {
  type: "STRING" || "NUMBER"
  scannerType: "CODE_39" || "QR"
  validation: "NRIC" || "REGEX"
  validationRegex?: <regex for voucher and QR>
}
```

- [x] add new scanner feature flag
- [x] use new flag for validation 
- [x] add regex validation option
- [x] refactor components and variables that use the term `NRIC` and replace with `id` to cater for future distributions using different identification

### DIFF
--- a/src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
+++ b/src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
@@ -25,6 +25,7 @@ import {
 import { IdScanner } from "../IdScanner/IdScanner";
 import { BarCodeScanner, BarCodeScannedCallback } from "expo-barcode-scanner";
 import { validateAndCleanNric } from "../../utils/validateNric";
+import { validateAndCleanId } from "../../utils/validateInputWithRegex";
 import { InputNricSection } from "./InputNricSection";
 import { AppHeader } from "../Layout/AppHeader";
 import * as Sentry from "sentry-expo";
@@ -36,6 +37,7 @@ import { ImportantMessageContentContext } from "../../context/importantMessage";
 import { useCheckUpdates } from "../../hooks/useCheckUpdates";
 import { KeyboardAvoidingScrollView } from "../Layout/KeyboardAvoidingScrollView";
 import { useProductContext } from "../../context/products";
+import { EnvVersionError } from "../../services/envVersion";
 
 const styles = StyleSheet.create({
   content: {
@@ -115,6 +117,9 @@ const CollectCustomerDetailsScreen: FunctionComponent<NavigationFocusInjectedPro
       setIsScanningEnabled(false);
       if (features?.id.validation === "NRIC") {
         nric = validateAndCleanNric(input);
+      } else if (features?.id.validation === "REGEX") {
+        const idRegex = features?.id.validationRegex;
+        nric = validateAndCleanId(input, idRegex);
       } else {
         // Remove validation
         nric = input;
@@ -124,6 +129,9 @@ const CollectCustomerDetailsScreen: FunctionComponent<NavigationFocusInjectedPro
       setNricInput("");
     } catch (e) {
       setIsScanningEnabled(false);
+      if (e instanceof EnvVersionError) {
+        Sentry.captureException(e);
+      }
       Alert.alert(
         "Error",
         e.message || e,
--- a/src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
+++ b/src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
@@ -186,9 +186,9 @@ const CollectCustomerDetailsScreen: FunctionComponent<NavigationFocusInjectedPro
           onCancel={() => setShouldShowCamera(false)}
           cancelButtonText="Enter ID manually"
           barCodeTypes={
-            features?.id.scannerType === "CODE_39"
-              ? [BarCodeScanner.Constants.BarCodeType.code39]
-              : [BarCodeScanner.Constants.BarCodeType.qr]
+            features?.id.scannerType === "QR"
+              ? [BarCodeScanner.Constants.BarCodeType.qr]
+              : [BarCodeScanner.Constants.BarCodeType.code39]
           }
         />
       )}
--- a/src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
+++ b/src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
@@ -24,8 +24,7 @@ import {
 } from "react-navigation";
 import { IdScanner } from "../IdScanner/IdScanner";
 import { BarCodeScanner, BarCodeScannedCallback } from "expo-barcode-scanner";
-import { validateAndCleanNric } from "../../utils/validateNric";
-import { validateAndCleanId } from "../../utils/validateInputWithRegex";
+import { validateAndCleanId } from "../../utils/validateIdentification";
 import { InputIdSection } from "./InputIdSection";
 import { AppHeader } from "../Layout/AppHeader";
 import * as Sentry from "sentry-expo";
@@ -112,21 +111,13 @@ const CollectCustomerDetailsScreen: FunctionComponent<NavigationFocusInjectedPro
   }, [shouldShowCamera]);
 
   const onCheck = async (input: string): Promise<void> => {
-    let id: string;
     try {
       setIsScanningEnabled(false);
-      switch (features?.id.validation) {
-        case "NRIC":
-          id = validateAndCleanNric(input);
-          break;
-        case "REGEX":
-          const idRegex = features?.id.validationRegex;
-          id = validateAndCleanId(input, idRegex);
-          break;
-        default:
-          // Remove validation
-          id = input;
-      }
+      const id = validateAndCleanId(
+        input,
+        features?.id.validation,
+        features?.id.validationRegex
+      );
       Vibration.vibrate(50);
       navigation.navigate("CustomerQuotaScreen", { id });
       setIdInput("");
--- a/src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
+++ b/src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
@@ -26,7 +26,7 @@ import { IdScanner } from "../IdScanner/IdScanner";
 import { BarCodeScanner, BarCodeScannedCallback } from "expo-barcode-scanner";
 import { validateAndCleanNric } from "../../utils/validateNric";
 import { validateAndCleanId } from "../../utils/validateInputWithRegex";
-import { InputNricSection } from "./InputNricSection";
+import { InputIdSection } from "./InputIdSection";
 import { AppHeader } from "../Layout/AppHeader";
 import * as Sentry from "sentry-expo";
 import { HelpButton } from "../Layout/Buttons/HelpButton";
@@ -70,7 +70,7 @@ const CollectCustomerDetailsScreen: FunctionComponent<NavigationFocusInjectedPro
   const messageContent = useContext(ImportantMessageContentContext);
   const [shouldShowCamera, setShouldShowCamera] = useState(false);
   const [isScanningEnabled, setIsScanningEnabled] = useState(true);
-  const [nricInput, setNricInput] = useState("");
+  const [idInput, setIdInput] = useState("");
   const { config } = useConfigContext();
   const showHelpModal = useContext(HelpModalContext);
   const checkUpdates = useCheckUpdates();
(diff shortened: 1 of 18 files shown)


src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
```diff
@@ -108,14 +112,29 @@ const CollectCustomerDetailsScreen: FunctionComponent<NavigationFocusInjectedPro
...
+          id = validateAndCleanNric(input);
+          break;
+        case "REGEX":
+          const idRegex = features?.id.validationRegex;
+          id = validateAndCleanId(input, idRegex);
+          break;
+        default:
+          // Remove validation
+          id = input;
+      }
```
sebastianquek, line 129: Shall we move this to a generic `validateAndCleanId` utility instead?

src/utils/validateInputWithRegex.test.tsx
```diff
@@ -0,0 +1,30 @@
+import { validate, validateAndCleanId } from "./validateInputWithRegex";
+import { EnvVersionError } from "../services/envVersion";
+
+describe("validate", () => {
+  const alphanumericRegex = "^[a-zA-Z0-9-_ ]+$";
```
sebastianquek, line 5: This example regex also includes `-`, `_` and ` `, do we want to test for these too?

src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
dalsontws: yep! i think we can do that, good to move this out too actually, since `AddUserModal` has the same chunk of code😛

src/utils/validateIdentification.ts
```diff
@@ -0,0 +1,27 @@
+import { validateAndCleanNric } from "./validateNric";
+import { validateAndCleanRegexInput } from "./validateInputWithRegex";
+import { EnvVersionError } from "../services/envVersion";
+
+export const validateAndCleanId = (
+  inputId: string,
+  idValidation: string | undefined,
+  idRegex: string | undefined
```
sebastianquek: Shall we set these as optional? So if consumers call `validateAndCleanId("ASD")` it works fine without needing the 2nd and 3rd args

dalsontws: eh i think i just added this liao 😂😂

src/types.ts
```diff
@@ -93,10 +93,22 @@ const Policy = t.intersection([
   })
 ]);
 
+const IdentificationFlag = t.intersection([
+  t.type({
+    type: t.string,
+    scannerType: t.string,
+    validation: t.string
```
sebastianquek, line 100: Do you think it's worthwhile to define this as a union of literals (i.e. enum)?

src/types.ts
```diff
@@ -93,10 +93,22 @@ const Policy = t.intersection([
   })
 ]);
 
+const IdentificationFlag = t.intersection([
+  t.type({
+    type: t.union([t.literal("STRING"), t.literal("NUMBER")]),
```
sebastianquek, line 98: Oh just noticed this,  we can use this to change the keyboard in the input id section

src/utils/validateIdentification.ts
sebastianquek: oh haha, how about `idValidation` too?

dalsontws: OH oops missed that, adding it in, thanks!🙏

src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
```diff
@@ -174,6 +185,11 @@ const CollectCustomerDetailsScreen: FunctionComponent<NavigationFocusInjectedPro
           onBarCodeScanned={onBarCodeScanned}
           onCancel={() => setShouldShowCamera(false)}
           cancelButtonText="Enter ID manually"
+          barCodeTypes={
+            features?.id.scannerType === "CODE_39"
+              ? [BarCodeScanner.Constants.BarCodeType.code39]
+              : [BarCodeScanner.Constants.BarCodeType.qr]
```
ilhaoni, line 191: Hmm I wonder if we should reverse this. Ideally our default should be code39 haha. What do you think haha

src/components/CustomerDetails/InputIdSection.tsx
```diff
@@ -39,18 +39,20 @@ const styles = StyleSheet.create({
   }
 });
 
-interface InputNricSection {
+interface InputIdSection {
```
ilhaoni, line 42: Hmm should we name the interface differently? Currently the function component below and this interface have the same name haha

src/components/CustomerQuota/AddUserModal.tsx
```diff
@@ -143,6 +149,11 @@ export const AddUserModal: FunctionComponent<AddUserModal> = ({
             onBarCodeScanned={onBarCodeScanned}
             onCancel={() => setShouldShowCamera(false)}
             cancelButtonText="Enter ID manually"
+            barCodeTypes={
```
ilhaoni, line 152: Hmm similar comment here, should default be code39 instead haha

src/components/CustomerQuota/AddUserModal.tsx
```diff
@@ -156,17 +167,18 @@ export const AddUserModal: FunctionComponent<AddUserModal> = ({
             <Card style={styles.card}>
               <View style={styles.cardHeader}>
                 <AppText style={{ flex: 1 }}>
-                  Add another NRIC to combine customer quotas
+                  Add another ID to combine customer quotas
```
ilhaoni, line 170: I feel like as an optimisation later on, we can dynamically word the type of id in our messages haha. Just one idea!

src/hooks/useCart/useCart.test.tsx
```diff
@@ -90,7 +90,12 @@ const defaultProducts: EnvVersion = {
   features: {
     REQUIRE_OTP: true,
     TRANSACTION_GROUPING: true,
-    FLOW_TYPE: "DEFAULT"
+    FLOW_TYPE: "DEFAULT",
+    id: {
```
ilhaoni, line 94: Lol super small nit, but the caps is different. HAHAHA. It's ok we can change later, not a must this PR 😂

src/services/envVersion/envVersion.test.tsx
```diff
@@ -43,7 +43,12 @@ const mockGetEnvVersionValidResponse = {
   features: {
     REQUIRE_OTP: true,
     TRANSACTION_GROUPING: true,
-    FLOW_TYPE: "DEFAULT"
+    FLOW_TYPE: "DEFAULT",
```
ilhaoni, line 46: Hmm, is there a reason why we initialise this multiple times ah. Rather than centralising it

src/utils/validateIdentification.test.ts
```diff
@@ -0,0 +1,52 @@
...
+    const idRegexValidation = "REGEX";
+    const alphanumericRegex = "^[a-zA-Z0-9-_ ]+$";
+    expect(
+      validateAndCleanId("abcd_1234", idRegexValidation, alphanumericRegex)
+    ).toBe("ABCD_1234");
+  });
+
+  it("should return existing input as ID if validation is missing", () => {
+    expect.assertions(1);
+    expect(validateAndCleanId("abcd_1234")).toBe("abcd_1234");
```
ilhaoni, line 22: Can I suggest to send a warning to Sentry when this happens? Ideally we'll always have validation, so this will let us know if validation is missing (: What do you think?

src/utils/validateInputWithRegex.test.tsx
```diff
@@ -0,0 +1,34 @@
+import { validate, validateAndCleanRegexInput } from "./validateInputWithRegex";
+import { EnvVersionError } from "../services/envVersion";
+
+describe("validate", () => {
+  const alphanumericRegex = "^[a-zA-Z0-9-_ ]+$";
+  it("should return true for alphanumeric regex", () => {
+    expect.assertions(4);
```
ilhaoni, line 7: Ooh interesting! What does this do haha

src/utils/validateInputWithRegex.ts
```diff
@@ -0,0 +1,26 @@
...
+  idRegex: string | undefined
+): string => {
+  if (!idRegex)
+    throw new EnvVersionError(
+      "Encountered an issue obtaining environment information. We've noted this down and are looking into it!"
+    );
+  const isValid = validate(inputId, idRegex);
+  if (!isValid)
+    throw new Error(
+      "Please check that the identification is in the correct format"
```
ilhaoni, line 18: So here we use the long form instead of id ah haha

src/utils/validateIdentification.test.ts
dalsontws: yes! will be ensuring `idValidation` is not undefined, and will throw EnvVersionError (which is caught in `CollectCustomerDetailsScreen` and send warning to sentry)

src/components/CustomerQuota/AddUserModal.tsx
dalsontws: previous Eida was working on standardization for all error messages to have generic words like "ID" instead of NRIC, should have ID here?

src/utils/validateInputWithRegex.ts
dalsontws: eh no lets do ID😅

src/hooks/useCart/useCart.test.tsx
dalsontws: 😂 yes it's quite a sore eye! i'll just change it in this PR bah? then remove the redundant `REQUIRE_OTP` flag as well!

dalsontws: HAHAA agree, super eye sore. I'll create a card for this changes + removing `REQUIRE_OTP` since its unused!

src/services/envVersion/envVersion.test.tsx
dalsontws: hmm this is for the test case, used to mockFetch the /version endpoint!

src/utils/validateInputWithRegex.test.tsx
dalsontws: this is used to ensure a certain number of assertions are called in a test case!😀

src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
dalsontws: agreed!

src/components/CustomerDetails/InputIdSection.tsx
dalsontws: hmm, looking thru our codebase, this seems like this is the usual practice, maybe @sebastianquek was there any reason why this was the case? i.e. interface and function having identical naming

src/hooks/useCart/useCart.test.tsx
sebastianquek: I think can leave it first? Once the campaign config parts are up, we'll use the normal camelCase convention.

src/components/CustomerDetails/InputIdSection.tsx
sebastianquek: Yea good question, I took reference from somewhere sometime back, but I can't remember where. Some considerations:
1. There's little use in differentiating between the interface and function for React function components.
Typescript and our IDEs already help to resolve whether we want to use a name as an interface or function. For example, let's say we export this `InputIdSection` interface as well. In other modules, we can do the following and everything works fine:
```ts
import { InputIdSection } from "./InputIdSection"; // Just need 1 import
...

interface ExtendedInputIdSection extends InputIdSection { // If the `InputIdSection` is not exported, then we'll immediately get a warning
  newProp: boolean;
}
...

const SomeComponent: FunctionComponent = () => (
  <InputIdSection ... /> // Resolved nicely here
)

```

2. We could add `Props` to the back of these interfaces, but those are additional constraints that we need to enforce. Right now we can do either an inline type, extracted type or extracted interface. So far this flexibility hasn't caused issues and the general rule of thumb is that for small functions with few props, just do it as an inline type. Am open to enforcing more constraints, but if we cannot automate and detect it, may be more of a bane.

src/components/CustomerQuota/AddUserModal.tsx
ilhaoni: ohh, that is true haha. Yeah let's leave it as ID then~

src/hooks/useCart/useCart.test.tsx
ilhaoni: haha okk~~ sounds good!

src/utils/validateInputWithRegex.test.tsx
ilhaoni: oooh, got it! haha til xD

src/utils/validateInputWithRegex.ts
ilhaoni: haha okk xD

src/services/envVersion/envVersion.test.tsx
ilhaoni: I see I see! okok haha

src/components/CustomerDetails/InputIdSection.tsx
ilhaoni: I see I see~ oooh is it this? 
https://www.typescriptlang.org/docs/handbook/declaration-merging.html

src/hooks/useCart/useCart.test.tsx
dalsontws: alright, shall leave it for now!

src/components/CustomerQuota/AddUserModal.tsx
dalsontws: agreed!

src/components/CustomerDetails/InputIdSection.tsx
sebastianquek: I think it's less about merging more of the name referring to different entities, perhaps more about polymorphism?

PR #105, opened 2020-07-17
feat: drawer nav + dynamic versioning

justussoh:
This PR includes:
 - adding of drawer navigation
 - version with jsBuildNumber


package.json
```diff
@@ -18,7 +18,7 @@
     "expo": "^35.0.0",
     "expo-barcode-scanner": "~7.0.0",
     "expo-camera": "^7.0.0",
-    "expo-constants": "^7.0.1",
+    "expo-constants": "~7.0.0",
```
sebastianquek, line 21: What's the rationale behind this change?

app.config.js
```diff
@@ -0,0 +1,8 @@
+export default ({ config }) => {
+  return {
+    ...config,
+    extra: {
+      appBuildVersion: process.env.APP_BUILD_VERSION
+    }
+  };
+};
```
sebastianquek, line 8: Is it possible to have it as `app.config.js`, so we minimise the use of `.js` files

src/components/Layout/DrawerNavigation.tsx
```diff
@@ -0,0 +1,196 @@
...
+  const showHelpModal = useContext(HelpModalContext);
+  const { drawerButtons } = useDrawerContext();
+  const handleLogout = useCallback((): void => {
+    logout(navigation.dispatch);
+  }, [logout, navigation.dispatch]);
+
+  const onPressLogout = (): void => {
+    Alert.alert(
+      "Are you sure you want to log out?",
+      "Stay logged in to change your scanning location.",
```
sebastianquek, line 102: Please remove this

src/components/Layout/DrawerNavigation.tsx
```diff
@@ -0,0 +1,196 @@
...
+  navLink: {
+    flexDirection: "row",
+    alignItems: "center",
+    paddingHorizontal: size(4),
+    paddingVertical: size(2.5),
+    borderBottomColor: color("grey", 20),
+    borderBottomWidth: 1
+  },
+  navLinkText: {
+    color: color("blue", 50),
```
sebastianquek, line 37: For all the default AppText, there's no need for this since they are blue 50 by default

src/components/Layout/AppHeader.tsx
```diff
@@ -26,46 +24,23 @@ export const AppHeaderComponent: FunctionComponent<AppHeader> = ({
...
       <AppName mode={mode} />
-      <TouchableOpacity onPress={onPressLogout}>
+      <TouchableOpacity onPress={onPressOpenDrawer}>
         <View
           style={{
-            flexDirection: "row",
-            alignItems: "center"
+            flexDirection: "row"
           }}
         >
```
sebastianquek, line 40: We don't need this wrapping view anymore given that there's only 1 child

src/context/drawer.tsx
```diff
@@ -0,0 +1,48 @@
+import React, {
+  createContext,
+  useContext,
+  FunctionComponent,
+  useState,
+  useCallback
+} from "react";
+
+export interface DrawerButtons {
```
sebastianquek, line 9: Just noticed this, should be `DrawerButton`, singular

package.json
justussoh: This was probably because we didn't know expo-constants was installed so ran `expo install` instead of npm, `expo install` grabs the correct version for the sdk from npm which explains the downgrade and constant version! should I discard this change?

app.config.js
justussoh: erm you mean `.ts`?

src/components/Layout/DrawerNavigation.tsx
```diff
@@ -0,0 +1,196 @@
...
+          Privacy Statement
+        </BottomNavigationLink>
+        <BottomNavigationLink
+          onPress={() => {
+            Linking.openURL("https://www.tech.gov.sg/report_vulnerability");
+          }}
+        >
+          Report vulnerability
+        </BottomNavigationLink>
+        <View style={{ marginTop: size(3), ...styles.bottomNavContainerLink }}>
```
sebastianquek, line 188: Can use a different style "class" instead, it isn't a "bottomNavContainerLink", also we don't need the view since we can style the AppText instead

src/components/Layout/DrawerNavigation.tsx
```diff
@@ -0,0 +1,196 @@
...
+          <AppText style={{ color: color("blue", 50), fontSize: fontSize(-4) }}>
+            {`Version: ${Constants.manifest.version}/ ${APP_BUILD_VERSION}`}
```
sebastianquek, line 190:
```suggestion
            {`Version: ${Constants.manifest.version} / ${APP_BUILD_VERSION}`}
```

src/components/Layout/DrawerNavigation.tsx
```diff
@@ -0,0 +1,196 @@
...
+  };
+
+  const onPressCloseDrawer = (): void => {
+    navigation.dispatch(DrawerActions.closeDrawer());
+  };
+
+  return (
+    <SafeAreaView
+      style={styles.container}
+      forceInset={{ top: "always", horizontal: "never" }}
```
sebastianquek, line 124: What's the purpose of this prop?

src/components/Layout/DrawerNavigation.tsx
```diff
@@ -0,0 +1,196 @@
...
+          top: size(5),
+          padding: size(1)
+        }}
+      >
+        <MaterialCommunityIcons
+          name="close"
+          size={size(3)}
+          color={color("blue", 50)}
+        />
+      </TouchableOpacity>
```
sebastianquek, line 140: I suggest we wrap this inside a `View` with a `marginTop`, so we don't need to absolute this button. The issue with absoluting is that if we choose to increase the `top` value, then we need to increase the sibling's `marginTop` too. Might as well rely on the normal document flow to push the subsequent elements down.

src/components/Layout/DrawerNavigation.tsx
```diff
@@ -0,0 +1,196 @@
...
+      <View
+        style={{
+          marginTop: size(9),
```
sebastianquek, line 143: Once the above is done, we can reduce this to maybe `size(3)` or `size(2)`, gotta see how it looks

src/components/Layout/DrawerNavigation.tsx
```diff
@@ -0,0 +1,196 @@
...
+          borderTopColor: color("grey", 20),
+          borderTopWidth: 1
+        }}
+      >
+        {drawerButtons.map(button => (
+          <DrawerButton {...button} key={button.label} />
+        ))}
+        <TouchableOpacity onPress={onPressLogout}>
+          <View style={styles.navLink}>
```
sebastianquek, line 152: Do we need this additional nested View?

sebastianquek: Can we make use of `DrawerButton` too?

app.config.js
sebastianquek: yep, sorry typo. If the typescript types arent stable, then we can just use `any`, but make sure to comment that once the types are ready, then we should add it

src/components/Layout/DrawerNavigation.tsx
justussoh: This is to ensure that the links don't get block by the notch at the top:
https://reactnavigation.org/docs/3.x/handling-iphonex/#use-forceinset-to-get-more-control

sebastianquek: By default doesn't the top inset always exist? Do we need the horizontal one given that the app only works in portrait?

src/components/Layout/DrawerNavigation.tsx
```diff
@@ -0,0 +1,195 @@
...
+    paddingVertical: size(1)
+  },
+  bottomVersionText: {
+    paddingHorizontal: size(4),
+    marginTop: size(3),
+    paddingVertical: size(1),
+    fontSize: fontSize(-4)
+  },
+  bottomNavContainerText: {
+    color: color("blue", 50),
```
sebastianquek, line 52: Missed out one :p

src/components/Layout/DrawerNavigation.tsx
```diff
@@ -0,0 +1,194 @@
...
+    navigation.dispatch(DrawerActions.closeDrawer());
+  };
+
+  return (
+    <SafeAreaView style={styles.container}>
+      <View
+        style={{
+          marginTop: size(4),
+          paddingRight: size(3),
+          flexDirection: "row-reverse"
```
sebastianquek, line 131:
```suggestion
          marginTop: size(5),
          marginRight: size(2),
          alignItems: "flex-end"
```

.github/workflows/deploy_branch_preview.yml
```diff
@@ -60,8 +60,21 @@ jobs:
...
+        id: latestRelease
+        uses: pozetroninc/github-action-get-latest-release@master
+        with:
+          repository: ${{ github.repository }}
+      - name: Echo Version Details
+        env:
+          LATEST_RELEASE: ${{steps.latestRelease.outputs.release}}
+        run: |
+          echo Build number is $GITHUB_RUN_NUMBER
+          echo Build number is ${LATEST_RELEASE#*-}
```
sebastianquek, line 73: Probably wanna change the output here

app.json
```diff
@@ -5,7 +5,6 @@
     "privacy": "public",
     "sdkVersion": "35.0.0",
     "platforms": ["ios", "android", "web"],
-    "version": "3.6.1",
```
sebastianquek, line 8: How about the `ios.buildNumber` and `android.versionCode`?

.github/workflows/deploy_branch_preview.yml
```diff
@@ -60,8 +60,21 @@ jobs:
             ${{ runner.os }}-
       - name: Install Packages
         run: npm install
+      - name: Get Latest Release
+        id: latestRelease
+        uses: pozetroninc/github-action-get-latest-release@master
+        with:
+          repository: ${{ github.repository }}
```
sebastianquek, line 67: Is it possible to get the latest tag by walking up the tree of the target branch instead? I doubt a global check of the latest release is scalable when we have multiple branches

.github/workflows/deploy_branch_preview.yml
```diff
@@ -60,8 +60,21 @@ jobs:
...
+        env:
+          LATEST_RELEASE: ${{steps.latestRelease.outputs.release}}
+        run: |
+          echo Build number is $GITHUB_RUN_NUMBER
+          echo Build number is ${LATEST_RELEASE#*-}
       - name: Expo Publish Channel
-        run: expo publish --non-interactive --release-channel pr${{ github.event.number }}
+        env:
+          LATEST_RELEASE: ${{steps.latestRelease.outputs.release}}
+        run: APP_BUILD_VERSION=$GITHUB_RUN_NUMBER VERSION=${LATEST_RELEASE#*-} expo publish --non-interactive --release-channel pr${{ github.event.number }}
```
sebastianquek, line 77: Can we use `APP_BINARY_VERSION` to correspond with the build version?

.github/workflows/deploy_branch_preview.yml
```diff
@@ -37,6 +37,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.head_ref }}
+      - name: Fetch Tags
+        run: |
+          git fetch --prune --unshallow --tags
```
sebastianquek, line 44: Hmm do we really need this?

app.config.ts
```diff
@@ -0,0 +1,18 @@
...
+    version: process.env.APP_BINARY_VERSION,
+    android: {
+      ...config.android,
+      versionCode: process.env.APP_BUILD_VERSION
+        ? parseInt(process.env.APP_BUILD_VERSION)
+        : 38
+    },
+    ios: {
+      ...config.ios,
+      buildNumber: process.env.APP_BUILD_VERSION
```
sebastianquek, line 15: what will this fallback to?

.github/workflows/deploy_branch_preview.yml
justussoh: Yes by default checkout v2 doesnt fetch the git tags so we cant find the latsest tag in the tree, i tried to fetch the tags in the best way possible

app.config.ts
justussoh: err in all the cd cases, there will never be any undefined for those env varibles. Might be a problem if someone is trying to manually deploy. Hmm we can possiblely add a default value! What do you suggest?

.github/workflows/deploy_branch_preview.yml
```diff
@@ -60,8 +65,21 @@ jobs:
...
       - name: Install Packages
         run: npm install
+      - name: Get Latest Release
+        id: latestRelease
+        run: |
+          RELEASE_TAG=$(echo $(git describe --tags --abbrev=0))
+          echo "::set-output name=release::$RELEASE_TAG"
+      - name: Echo Version Details
+        env:
+          LATEST_RELEASE: ${{steps.latestRelease.outputs.release}}
```
sebastianquek, line 75: Since we set it as an output in the previous step, do we need this `env`?

.github/workflows/deploy_branch_preview.yml
```diff
@@ -60,8 +65,21 @@ jobs:
             ${{ runner.os }}-
       - name: Install Packages
         run: npm install
+      - name: Get Latest Release
```
sebastianquek, line 68: Can we rename this step to be more explicit on its purpose, i.e. latest binary version

.github/workflows/deploy_branch_preview.yml
```diff
@@ -60,8 +65,21 @@ jobs:
...
+        id: latestRelease
+        run: |
+          RELEASE_TAG=$(echo $(git describe --tags --abbrev=0))
+          echo "::set-output name=release::$RELEASE_TAG"
```
sebastianquek, line 72: Since we output this directly, won't the `release` be `prod-x.x.x`, when we actually want `x.x.x`?

.github/workflows/deploy_branch_preview.yml
```diff
@@ -60,8 +65,21 @@ jobs:
...
+      - name: Echo Version Details
+        env:
+          LATEST_RELEASE: ${{steps.latestRelease.outputs.release}}
+        run: |
+          echo Build number is $GITHUB_RUN_NUMBER
+          echo Latest release is ${LATEST_RELEASE#*-}
       - name: Expo Publish Channel
-        run: expo publish --non-interactive --release-channel pr${{ github.event.number }}
+        env:
+          LATEST_RELEASE: ${{steps.latestRelease.outputs.release}}
```
sebastianquek, line 81: Same comment as above, do we need this?

.github/workflows/deploy_prod.yml
```diff
@@ -46,8 +47,21 @@ jobs:
           echo "::set-output name=releaseChannel::$RELEASE_CHANNEL"
       - name: Install Packages
         run: npm install
+      - name: Get Latest Release # Release tag will be prod-<version> i.e. prod-3.7.1, prod-3.8.0
+        id: latestRelease
+        uses: pozetroninc/github-action-get-latest-release@master
+        with:
+          repository: ${{ github.repository }}
```
sebastianquek, line 54: Hmm can we standardise with the branch preview one? Realised that in future when devs see this, they will likely be thinking why the difference.

app.config.ts
```diff
@@ -0,0 +1,18 @@
...
+// Install when on stable release (https://www.npmjs.com/package/@expo/config)
+export default ({ config }: any): any => {
+  return {
+    ...config,
+    version: process.env.APP_BINARY_VERSION,
+    android: {
+      ...config.android,
+      versionCode: process.env.APP_BUILD_VERSION
+        ? parseInt(process.env.APP_BUILD_VERSION)
+        : 38
```
sebastianquek, line 11: Let's move this to a constant so that we can describe what this magic number actually is

app.config.ts
sebastianquek: As you mentioned about the error, can we perform the check and throw appropriate errors? So when devs run it locally, they will know why it's failing.

.github/workflows/deploy_branch_preview.yml
justussoh: erm im using bash's param substitution to remove the `prod-` which only works if its a bash env variable so I'm injecting that. alternatively, i can shift that into the run script

.github/workflows/deploy_branch_preview.yml
justussoh: hmmm true

.github/workflows/deploy_branch_preview.yml
```diff
@@ -65,10 +65,12 @@ jobs:
             ${{ runner.os }}-
       - name: Install Packages
         run: npm install
-      - name: Get Latest Binary Version #Binary Version will be x.x.x based on the latest tag
+      - name: Get Latest Binary Version # Binary Version will be x.x.x based on the latest tag
         id: latestBinaryVersion
         run: |
+          # Release tag finds the lastest tag in the tree branch - i.e. prod-x.x.x
           RELEASE_TAG=$(echo $(git describe --tags --abbrev=0))
+          # Using param subsition, we output x.x.x instead
```
sebastianquek, line 73: Typo here haha

src/components/Layout/DrawerNavigation.tsx
```diff
@@ -0,0 +1,205 @@
...
+    }
+  } else if (__DEV__) {
+    version += "Dev version";
+  }
+
+  return (
+    <SafeAreaView style={styles.container}>
+      <View
+        style={{
+          marginTop: size(5),
```
sebastianquek, line 142:
```suggestion
          marginTop: size(8),
```

PR #106, opened 2020-07-20
feat: scanning limit

sebastianquek:
Blocked by #107 

* This pertains to the merchant flow
* Sets a limit at maximum 200 vouchers per transaction

https://www.notion.so/Limit-to-max-200-vouchers-scanned-1927dbbdd886469995727081816b49d7

### DIFF
--- a/src/utils/validateVoucherCode.test.tsx
+++ b/src/utils/validateVoucherCode.test.tsx
@@ -33,10 +33,11 @@ describe("check duplicate", () => {
 
 describe("check limit reached", () => {
   it("should return true when number of voucher codes is above or equal to the limit", () => {
-    expect.assertions(3);
+    expect.assertions(4);
     expect(checkLimitReached(["1", "2"], 2)).toBe(true);
     expect(checkLimitReached(["1", "2", "3"], 2)).toBe(true);
     expect(checkLimitReached(["1", "2", "3", "4"], 2)).toBe(true);
+    expect(checkLimitReached(["a", "b", "c", "d"], 3)).toBe(true);
   });
 
   it("should return false when number of voucher codes is below the limit", () => {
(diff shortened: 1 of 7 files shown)


src/utils/validateVoucherCode.test.tsx
```diff
@@ -26,3 +30,17 @@ describe("check duplicate", () => {
...
+    expect.assertions(2);
+    expect(checkLimitReached(3, 2)).toBe(true);
+    expect(checkLimitReached(4, 2)).toBe(true);
+  });
+
+  it("should return false when number of voucher codes is below the limit", () => {
+    expect.assertions(2);
+    expect(checkLimitReached(0, 2)).toBe(false);
+    expect(checkLimitReached(2, 2)).toBe(false);
+  });
```
dalsontws, line 45: Would testing negative numbers be helpful? not sure if that could happen haha

src/utils/validateVoucherCode.tsx
```diff
@@ -11,15 +12,36 @@ export const checkDuplicate = (
...
+  limit = maxNumVouchersPerTransaction
+): boolean => {
+  return lengthOfVoucherCodes > limit;
+};
+
 export const validateVoucherCode = (
   voucherCode: string,
   voucherCodes: string[]
 ): boolean => {
+  const hasReachedLimit = checkLimitReached(voucherCodes.length + 1);
```
dalsontws, line 33: just curious, any reason to use this method instead of having `lengthOfVoucherCodes >= limit` ?

sebastianquek: I've updated the code to use this instead, I initially did something similar, but somehow changed it to send in the length of the voucher codes 🤷

PR #107, opened 2020-07-20
chore: merchant flow dialog standardisation

sebastianquek:
https://www.notion.so/Merchant-flow-dialogs-e7ec131e72cc40e0abab034ad8397e02

### DIFF
--- a/src/components/MerchantPayout/VoucherStatusModal/VoucherStatusModal.tsx
+++ b/src/components/MerchantPayout/VoucherStatusModal/VoucherStatusModal.tsx
@@ -1,4 +1,4 @@
-import React, { FunctionComponent } from "react";
+import React, { FunctionComponent, useEffect } from "react";
 import {
   View,
   StyleSheet,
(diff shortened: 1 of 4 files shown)


src/components/MerchantPayout/VoucherStatusModal/VoucherStatusModal.tsx
```diff
@@ -132,5 +145,5 @@ export const VoucherStatusModal: FunctionComponent<VoucherStatusModal> = ({
       <View style={styles.background} />
       <View style={styles.cardWrapper}>{card}</View>
     </Modal>
-  );
+  ) : null;
```
dalsontws, line 148: noticed an issue when manually keying in voucher codes, `ManualAddVoucherModal` cannot be closed or used to add any other codes

sebastianquek:
There's a very annoying bug where the React Native Modal and Alerts don't play well together:
https://stackoverflow.com/questions/40041472/react-native-modal-does-not-hide-when-alert-appears-on-top-of-it
https://stackoverflow.com/questions/53533750/popping-up-a-alert-in-front-of-react-native-modal-hangs-up-the-application-in-io

I've fixed some other issues and added some useCallbacks to see if reducing certain re-renders solves the problem. But this didn't change much, the issue is still apparent. I have resorted to using a double requestAnimationFrame to ensure the rendering of the alert is delayed. The above links all use a fixed timeout, but requestAnimationFrame. is more precise and just waits for the next animation frame.

PR #110, opened 2020-07-21
feat: use queryparams from deeplinks

sebastianquek:
Use this to generate the URLs: https://kitkm.csb.app?releaseChannel=pr110


younglim:
Deep links working on Android

PR #112, opened 2020-07-23
fix: bug fix for nric validation

dhemutton:
This PR contains a frontend fix for the NRIC regex. Previously was accepting NRIC with more than 9 digits.
Screenshot 2020-07-23 at 11 49 08 AM
Screenshot 2020-07-23 at 11 49 03 AM

### DIFF
--- a/src/utils/validateNric.tsx
+++ b/src/utils/validateNric.tsx
@@ -1,6 +1,6 @@
 // Thanks to https://gist.github.com/kyrene-chew/6f275325335ab27895beb7a9a7b4c1cb
 
-export const nricRegex = /(\D)(\d{7})(\D)/;
+export const nricRegex = /(\D)(\d{7})(\D)$/;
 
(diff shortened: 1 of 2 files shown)


src/utils/validateNric.tsx
```diff
@@ -1,6 +1,6 @@
 // Thanks to https://gist.github.com/kyrene-chew/6f275325335ab27895beb7a9a7b4c1cb
 
-export const nricRegex = /(\D)(\d{7})(\D)/;
+export const nricRegex = /(\D)(\d{7})(\D)$/;
```
Gabrieltay, line 3: Perhap should include `^`, else 12345S8888888W will be accepted too
```js
export const nricRegex = /^(\D)(\d{7})(\D)$/`;
```

dhemutton: ohyeahh thanks nice catch!

PR #114, opened 2020-07-24
feat: add ability to force check updates

sebastianquek:
Right now updates can only be checked every 60 minutes, there are situations where we will want to perform checks that are not constrained by this rule.

This PR also adds a return type so consumers can know the outcome of the check.

### DIFF
--- a/src/hooks/useCheckUpdates.tsx
+++ b/src/hooks/useCheckUpdates.tsx
@@ -15,7 +15,7 @@ type CheckUpdatesResult =
   | "UPDATE_ERROR";
 
 export const useCheckUpdates = (): ((
-  forceCheck: boolean
+  forceCheck?: boolean
 ) => Promise<CheckUpdatesResult>) => {
   const setMessageContent = useContext(ImportantMessageSetterContext);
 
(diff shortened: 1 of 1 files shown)


src/hooks/useCheckUpdates.tsx
```diff
@@ -8,53 +8,72 @@ import { ImportantMessageSetterContext } from "../context/importantMessage";
 const LAST_UPDATE_KEY = "LAST_UPDATE";
 const CHECK_UPDATE_INTERVAL_MINUTES = 60;
 
-export const useCheckUpdates = (): (() => void) => {
+type CheckUpdatesResult =
+  | "TOO_RECENT_CHECK"
```
ilhaoni, line 12: hahaha so `TOO_RECENT_CHECK` is like a default fallback value ah

sebastianquek: Hmmm, not really, the branch that leads to `TOO_RECENT_CHECK` occurs only when the last update was within 60mins.

sebastianquek: That reminds me, we wanted to add a condition to check every 30mins, i'll update this

ilhaoni: haha ok!

PR #115, opened 2020-07-24
feat: campaign config service

sebastianquek:
* Added a `campaignConfig` service that calls `${endpoint}/client-config` to retrieve the config.
* Added a `campaignConfig` context that allows consumers to retrieve the campaign config anywhere in the tree. This context exposes a `configHashes` that is updated every time a new configuration is saved and when it's loaded from async storage. This `configHashes` is used when calling the endpoint so that the endpoint does not need to return the same set of data if the client already has the most updated config

### DIFF
--- a/src/services/campaignConfig/campaignConfig.ts
+++ b/src/services/campaignConfig/campaignConfig.ts
@@ -1,5 +1,5 @@
 import { IS_MOCK } from "../../config";
-import { CampaignConfig } from "../../types";
+import { CampaignConfig, ConfigHashes } from "../../types";
 import { fetchWithValidator, ValidationError } from "../helpers";
 import * as Sentry from "sentry-expo";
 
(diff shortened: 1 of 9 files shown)


src/services/campaignConfig/campaignConfig.ts
```diff
@@ -0,0 +1,54 @@
...
+const mockGetCampaignConfig = async (
+  _token: string,
+  _endpoint: string,
+  configHashes: ConfigHashes
+): Promise<CampaignConfig> => {
+  return {
+    features: {
+      minAppBinaryVersion: "3.0.0",
+      minAppBuildVersion: 0
+    }
```
dalsontws, line 48: we will be having the existing features in campaign config as well right? i.e. transactionGrouping, id, flowType. 
or for now we are just using this endpoint to fetch the app versions?

sebastianquek: Yep we will add them in once we actually switch over, so yea you're right, just an endpoint to fetch the app versions

PR #117, opened 2020-07-24
feat: check version hook

sebastianquek:
This hook gets the version from `config` and compares it with the minimum version from the campaign config

Checks:
1. `config.APP_BINARY_VERSION` vs `features.minAppBinaryVersion`
2. `config.APP_BUILD_VERSION` vs `features.minAppBuildVersion`

**NOTE**
Location of the current app version is subject to change, depends on the navigation sidebar PR that configure the app version variables

### DIFF
--- a/src/hooks/useCheckVersion/useCheckVersion.ts
+++ b/src/hooks/useCheckVersion/useCheckVersion.ts
@@ -1,0 +1,60 @@
+import { useContext } from "react";
+import { CampaignConfigContext } from "../../context/campaignConfig";
+import * as config from "../../config";
+
+type CheckVersionResult = "OK" | "OUTDATED_BINARY" | "OUTDATED_BUILD";
+
+export const useCheckVersion = (): (() => CheckVersionResult) => {
+  const { features } = useContext(CampaignConfigContext);
+
+  const isBinaryVersionOK = (): boolean => {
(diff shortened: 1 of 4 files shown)


src/hooks/useCheckVersion/useCheckVersion.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    const currentBinaryVersionSplit = currentBinaryVersion
+      .split(".")
+      .map(Number);
+    const minBinaryVersionSplit = minBinaryVersion.split(".").map(Number);
+    for (let i = 0; i < minBinaryVersionSplit.length; i++) {
+      if (currentBinaryVersionSplit[i] < minBinaryVersionSplit[i]) {
+        return false;
+      } else if (currentBinaryVersionSplit[i] > minBinaryVersionSplit[i]) {
+        return true;
```
dalsontws, line 29: just clarifying, where will the binary version will be documented? (for us to track haha)

sebastianquek: Oh we can look at the releases on GitHub, each release creates a new binary version and build version, and also immediately publishes on the corresponding release channel.

However, we don't need to upload a new binary on every single release since the OTA from the publishing will provide the latest build version.

PR #121, opened 2020-07-27
refactor: move sentry out to its own module 

sebastianquek:
where we mock it on dev so we can see what is being called during development

### DIFF
--- a/src/utils/errorTracking.ts
+++ b/src/utils/errorTracking.ts
@@ -1,0 +1,29 @@
+import * as SentryExpo from "sentry-expo";
+
+const SENTRY_LOG_PREFIX = "[dev-sentry]";
+
+// This dev mock only implements the 4 currently used Sentry functions
+let Sentry: Pick<
+  typeof SentryExpo,
+  "init" | "setRelease" | "addBreadcrumb" | "captureException"
+>;
+
+if (__DEV__) {
+  Sentry = {
+    init: (): void => console.log(`${SENTRY_LOG_PREFIX} initialised`),
+    setRelease: (release: string): void =>
+      console.log(`${SENTRY_LOG_PREFIX} setRelease: ${release}`),
+    addBreadcrumb: (breadcrumb: unknown): void =>
+      console.log(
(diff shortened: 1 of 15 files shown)


src/utils/errorTracking.ts
```diff
@@ -0,0 +1,29 @@
...
+
+// This dev mock only implements the 4 currently used Sentry functions
+let Sentry: Pick<
+  typeof SentryExpo,
+  "init" | "setRelease" | "addBreadcrumb" | "captureException"
+>;
+
+if (__DEV__) {
+  Sentry = {
+    init: (): void => console.log(`${SENTRY_LOG_PREFIX} initialised`),
```
justussoh, line 13: Does this need to take in 1 param?

sebastianquek: The type is inherited from `sentry-expo`, so we don't need to worry about not specifying it here.

Since there's no important information in `init` that will be helpful when we debug, I've chosen to exclude it.

src/utils/errorTracking.ts
```diff
@@ -0,0 +1,29 @@
...
+      console.log(
+        `${SENTRY_LOG_PREFIX} addBreadcrumb: ${JSON.stringify(breadcrumb)}`
+      ),
+    captureException: (e: unknown): string => {
+      console.warn(`${SENTRY_LOG_PREFIX} captureException: ${e}`);
+      return "eventId";
+    }
+  };
+} else {
+  Sentry = SentryExpo;
```
ilhaoni, line 26: Hmm, are we just initialising it for dev first to try? 

Just curious why we're not configuring more if its not dev HAHA

sebastianquek: Hmm, not very sure what you mean. The goal of this abstraction is to make it easier to see which Sentry function is being called during development, and also to avoid issues where an error pops up saying that Sentry is not initialised (when it doesn't matter in dev).

If you're referring to the configuration, we initialise it in `App.tsx`, other than that no other configuration is needed.

ilhaoni: Ahh, okok, got it! :D Thanks for explaining haha

PR #122, opened 2020-07-27
feat: use expo-updates from sdk38

dhemutton:
This PR contains the implementation for expo-updates from SDK 38.


package.json
```diff
@@ -39,7 +39,8 @@
     "react-native-webview": "^10.3.2",
     "react-navigation": "^4.1.0",
     "react-navigation-stack": "^1.10.3",
-    "sentry-expo": "^2.0.3"
+    "sentry-expo": "^2.0.3",
+    "expo-updates": "~0.0.1-rc.0"
```
sebastianquek, line 43: Why is it on this weird release candidate?

src/hooks/useCheckUpdates.tsx
```diff
@@ -1,4 +1,4 @@
-import { Updates } from "expo";
+import * as Updates from "expo-updates";
```
sebastianquek, line 1: I think there's another place that uses this too?

package.json
dhemutton: hmm..i think it's still a beta version but it should be the latest cos i ran `expo install expo-updates` again and it results in the same version

PR #125, opened 2020-07-27
fix: sync build number

justussoh:
We need to synchronize the buildNumber and `GITHUB_RUN_NUMBER` before building and releasing it to the app stores.
Also, added `appBuildVersion` into extra for easy consumption

### DIFF
--- a/app.config.ts
+++ b/app.config.ts
@@ -12,8 +12,7 @@ const buildNumber: string = getValue(
   "Please provide a Build Version with APP_BUILD_VERSION env variable"
 );
 // The APP_BUILD_VERSION number is 38 behind the actual deploy, adding 38 synchronises the build number
-const androidVersionCode = parseInt(buildNumber) + 38;
-const appBuildVersion = androidVersionCode.toString();
+const appBuildVersion = parseInt(buildNumber) + 38;
 
 export default ({ config }: any): any => {
   return {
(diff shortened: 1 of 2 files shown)


app.config.ts
```diff
@@ -26,11 +28,12 @@ export default ({ config }: any): any => {
...
     ios: {
       ...config.ios,
-      buildNumber
+      buildNumber: appBuildVersion
     },
     extra: {
       mock: process.env.MOCK === "true",
-      storybook: process.env.START_STORYBOOK === "true"
+      storybook: process.env.START_STORYBOOK === "true",
+      appBuildVersion
```
sebastianquek, line 36: appBuildVersion should be a number

app.config.ts
```diff
@@ -11,7 +11,9 @@ const buildNumber: string = getValue(
   process.env.APP_BUILD_VERSION,
   "Please provide a Build Version with APP_BUILD_VERSION env variable"
 );
-const androidVersionCode: number = parseInt(buildNumber);
+// The APP_BUILD_VERSION number is 38 behind the actual deploy, adding 38 synchronises the build number
+const androidVersionCode = parseInt(buildNumber) + 38;
+const appBuildVersion = androidVersionCode.toString();
```
sebastianquek, line 16: This part's a bit clunky, I suggest to name it as the following, then only in the `ios.buildNumber` you directly call `.toString`
```suggestion
const appBuildVersion = parseInt(buildNumber) + 38;
```

PR #128, opened 2020-07-28
feature: alert modals for different severity

Gabrieltay:
- Added to Storybook
- `AlertType` eg. `ERROR` , `WARN`, `INFO` and `CONFIRM`
- AlertModal to be shown by using `showAlert` of `AlertModalContext`


src/context/alert.tsx
```diff
@@ -0,0 +1,62 @@
...
+    >
+      {children}
+      <AlertModal
+        alertType={alertProps.alertType}
+        title={alertProps.title}
+        description={alertProps.description}
+        buttonTextType={alertProps.buttonTextType}
+        visible={alertProps.visible}
+        onOk={alertProps.onOk}
+        onCancel={alertProps.onCancel}
```
sebastianquek, line 57: Would doing this be cleaner?
```suggestion
        {...alertProps}
```

src/context/alert.tsx
```diff
@@ -0,0 +1,62 @@
...
+});
+
+export const AlertModalContextProvider: FunctionComponent = ({ children }) => {
+  const [alertProps, setAlertProps] = useState<AlertModalProp>(
+    defaultAlertProp
+  );
+
+  const showAlert: AlertModalContext["showAlert"] = (props: AlertModalProp) => {
+    setAlertProps(props);
+  };
```
sebastianquek, line 36: Probably want to wrap these callbacks in `useCallback` so they don't cause unnecessary rerenders. Will need to take a closer look at performance in future

src/components/AlertModal/AlertModal.tsx
```diff
@@ -0,0 +1,165 @@
...
+import { DangerButton } from "../Layout/Buttons/DangerButton";
+import { size } from "../../common/styles";
+
+const styles = StyleSheet.create({
+  centeredView: {
+    flex: 1,
+    width: "100%",
+    justifyContent: "center",
+    alignItems: "center",
+    alignSelf: "center",
```
sebastianquek, line 15:
```suggestion
```

src/components/AlertModal/AlertModal.tsx
```diff
@@ -0,0 +1,165 @@
...
+    backgroundColor: "rgba(0,0,0,0.6)"
+  },
+  modalView: {
+    marginBottom: 0,
+    paddingTop: 24,
```
sebastianquek, line 20: We have a set of styling utilities that define the various sizes, so in this case it should be `size(3)`

src/components/AlertModal/AlertModal.tsx
```diff
@@ -0,0 +1,165 @@
...
+    width: 280,
+    backgroundColor: "white",
+    borderRadius: size(1),
```
sebastianquek, line 23: There's `borderRadius(n)` too

src/components/AlertModal/AlertModal.tsx
```diff
@@ -0,0 +1,165 @@
...
+    shadowColor: "#000",
+    shadowOffset: {
+      width: 0,
+      height: 2
+    },
+    shadowOpacity: 0.25,
+    shadowRadius: 3.84,
+    elevation: 5
```
sebastianquek, line 31: Can use the `shadow` util: `...shadow(2),`, gotta experiment with which level to use

src/components/AlertModal/AlertModal.tsx
```diff
@@ -0,0 +1,165 @@
...
+        );
+    }
+  }, [callToAction.primaryActionText, props]);
+
+  return (
+    <Modal animationType="fade" transparent={true} visible={props.visible}>
+      <View style={styles.centeredView}>
+        <View style={styles.modalView}>
+          {props.alertType === "ERROR" && (
+            <AlertLogo style={styles.alertIcon} />
```
sebastianquek, line 142: `AlertIcon` instead?

src/components/AlertModal/AlertModal.tsx
```diff
@@ -0,0 +1,165 @@
...
+    flexDirection: "column",
+    justifyContent: "center",
+    alignItems: "center",
+    alignSelf: "center",
+    marginBottom: 9
+  },
+  modalTitle: {
+    fontWeight: "bold",
+    fontFamily: "brand-bold",
+    fontSize: 20,
```
sebastianquek, line 43: There's the `fontSize(n)` helper too

src/components/AlertModal/AlertModal.tsx
```diff
@@ -0,0 +1,165 @@
...
+      height: 2
+    },
+    shadowOpacity: 0.25,
+    shadowRadius: 3.84,
+    elevation: 5
+  },
+  alertIcon: {
+    flexDirection: "column",
+    justifyContent: "center",
+    alignItems: "center",
```
sebastianquek, line 36: Don't need these styles since they don't affect anything
```suggestion
```

src/components/AlertModal/AlertModal.tsx
```diff
@@ -0,0 +1,165 @@
...
+    alignSelf: "center",
```
sebastianquek, line 37: Given that the entire modal is center aligned, can we do a `alignItems: "center"` on the `modalView` instead? Then  we don't need to specify `alignSelf` and only the description text needs `textAlign: "center"`

src/components/AlertModal/AlertModal.tsx
```diff
@@ -0,0 +1,165 @@
...
+    textAlign: "center",
+    fontSize: 16,
+    color: "#305367"
+  },
+  modalGroupButton: {
+    flexDirection: "row",
+    marginHorizontal: 16,
+    marginBottom: 24
+  },
+  modalSecondaryBtm: {
```
sebastianquek, line 61:
```suggestion
  modalSecondaryButton: {
```

src/components/AlertModal/AlertModal.tsx
```diff
@@ -0,0 +1,165 @@
...
+          <Text style={styles.modalText}>{props.description}</Text>
+          <View style={styles.modalGroupButton}>
+            {props.alertType !== "ERROR" && props.alertType !== "INFO" && (
+              <View style={styles.modalSecondaryBtm}>
+                <SecondaryButton
+                  text={callToAction.secondaryActionText || ""}
+                  fullWidth={true}
+                  onPress={() => {
+                    props.onExit();
+                    props.onCancel && props.onCancel();
```
sebastianquek, line 154: Some new typescript shortcuts
```suggestion
                    props.onCancel?.();
```

src/components/AlertModal/AlertModal.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+export interface AlertModalProp {
+  alertType: AlertType;
+  title: string;
+  description: string;
+  buttonTextType: CallToActionKey;
+  visible: boolean;
+  onOk: () => void;
+  onCancel?: () => void;
+  onExit?: any;
```
sebastianquek, line 98: Any reason for the any here?

src/components/AlertModal/AlertModal.tsx
```diff
@@ -0,0 +1,165 @@
...
+    secondaryActionText: "CANCEL"
+  }
+});
+
+interface CallToActionButtonTexts {
+  primaryActionText: string;
+  secondaryActionText?: string;
+}
+
+export type CallToActionKey = keyof typeof callToActionCollection;
```
sebastianquek, line 88: Interesting!

What's the reason for deriving the `CallToActionKey` from the implementation, rather than the inverse (i.e. define the `CallToActionKey`, then `callToActionCollection` adheres to the type by ensuring those keys are defined? I'm thinking we probably want then for different languages, what do you think?

src/components/AlertModal/AlertModal.tsx
```diff
@@ -0,0 +1,165 @@
...
+}
+
+export const AlertModal: FunctionComponent<AlertModalProp> = (
+  props: AlertModalProp
+) => {
+  const callToAction = useMemo(
+    () =>
+      callToActionCollection[props.buttonTextType] as CallToActionButtonTexts,
+    [props.buttonTextType]
+  );
```
sebastianquek, line 108: I noticed a few usages of `useMemo`, could you share more about why it's needed here?

src/components/AlertModal/AlertModal.tsx
```diff
@@ -0,0 +1,165 @@
...
+  description: string;
+  buttonTextType: CallToActionKey;
+  visible: boolean;
+  onOk: () => void;
+  onCancel?: () => void;
+  onExit?: any;
+}
+
+export const AlertModal: FunctionComponent<AlertModalProp> = (
+  props: AlertModalProp
```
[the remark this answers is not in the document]
Gabrieltay: sure!

src/components/AlertModal/AlertModal.tsx
```diff
@@ -0,0 +1,165 @@
...
+});
+
+interface CallToActionButtonTexts {
+  primaryActionText: string;
+  secondaryActionText?: string;
+}
+
+export type CallToActionKey = keyof typeof callToActionCollection;
+
+export interface AlertModalProp {
```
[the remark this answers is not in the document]
Gabrieltay: 👌

src/components/AlertModal/AlertModal.tsx
```diff
@@ -0,0 +1,165 @@
...
+  return (
+    <Modal animationType="fade" transparent={true} visible={props.visible}>
+      <View style={styles.centeredView}>
+        <View style={styles.modalView}>
+          {props.alertType === "ERROR" && (
+            <AlertLogo style={styles.alertIcon} />
+          )}
+          <Text style={styles.modalTitle}>{props.title}</Text>
+          <Text style={styles.modalText}>{props.description}</Text>
+          <View style={styles.modalGroupButton}>
```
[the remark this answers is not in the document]
Gabrieltay: Cool! Updated

src/components/AlertModal/AlertModal.tsx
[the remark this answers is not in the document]
Gabrieltay: Done

src/components/AlertModal/AlertModal.tsx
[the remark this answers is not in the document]
Gabrieltay: Done

src/context/alert.tsx
Gabrieltay: Good suggestion. Updated

src/components/AlertModal/AlertModal.tsx
Gabrieltay: good catch. Not sure why this was added in the first place. Updated

src/components/AlertModal/AlertModal.tsx
Gabrieltay: Updated

src/components/AlertModal/AlertModal.tsx
Gabrieltay: Updated

src/components/AlertModal/AlertModal.tsx
Gabrieltay: updated to `...shadow(1)`. To be honest, I can't really tell the difference  😅

src/components/AlertModal/AlertModal.tsx
Gabrieltay: Sure

src/components/AlertModal/AlertModal.tsx
Gabrieltay: Updated

src/components/AlertModal/AlertModal.tsx
Gabrieltay: You are right. Removed

src/components/AlertModal/AlertModal.tsx
Gabrieltay: Make sense. Updated

src/components/AlertModal/AlertModal.tsx
Gabrieltay: Updated

src/components/AlertModal/AlertModal.tsx
Gabrieltay: Niceeee 👍 . Updated

src/components/AlertModal/AlertModal.tsx
Gabrieltay: 🤔  doubt so. Updated with `() => void`

src/components/AlertModal/AlertModal.tsx
Gabrieltay: similar concept as `useCallback`. But now i think again whether its necessary to "memorize" function or value for such a transient component. What do you think?

src/components/AlertModal/AlertModal.tsx
Gabrieltay: Maybe you enlighten me on this. I have tried to declare the object by using the following
```ts
const callToActionCollection: { [key:"YES_NO" | "OK_CANCEL" | "CONFIRM_CANCEL"]: CallToActionButtonTexts }
```
But the `key` was expected to be `string` or `number`, not string literals.

Alternatively is to make use of `Map<"YES_NO" | "OK_CANCEL" | "CONFIRM_CANCEL" , CallToActionButtonTexts>`

How do you intend to cater for different languages?

src/context/alert.tsx
Gabrieltay: Just thinking that this is a transient component. Is it necessary to use `useCallback`?

src/components/AlertModal/AlertModal.tsx
sebastianquek: Yea, doesn't really matter since it's on a dark background anyway

src/components/AlertModal/AlertModal.tsx
sebastianquek: Just tried this approach:
```
type CallToActionKey = "YES_NO" | "OK_CANCEL" | "CONFIRM_CANCEL";
type CallToActionButtonTexts = {
  primaryActionText: string;
  secondaryActionText?: string;
};
type CallToActionCollection = {
  [key in CallToActionKey]: CallToActionButtonTexts;
};

const callToActionCollection: CallToActionCollection = {
  YES_NO: { primaryActionText: "YES", secondaryActionText: "NO" },
  OK_CANCEL: { primaryActionText: "OK", secondaryActionText: "CANCEL" },
  CONFIRM_CANCEL: {
    primaryActionText: "CONFIRM",
    secondaryActionText: "CANCEL"
  }
};
```

For the different languages, I'm thinking they can be supplied by the campaign configuration context. Then `AlertModal` can directly access the context to find the appropriate mappings for `callToActionCollection`. Similar to a `ThemeProvider` in CSS-in-JS libraries.

src/components/AlertModal/AlertModal.tsx
sebastianquek: Yea there are overheads with using `useCallback` and `useMemo`, perhaps we can defer the use of this and measure first? These components are afterall relatively small and aren't computationally expensive, and they don't trigger many other re-renders.

I had another comment about the contexts using `useCallback`s, I defaulted to ensuring context values are wrapped properly because a re-render in a context triggers many re-renders all the way down the tree. Though I haven't benchmarked or measured the perf yet, I've encountered some issues with context re-rendering too many times

src/context/alert.tsx
sebastianquek: Hmm, my main concern/hypothesis is when any of its parent context providers re-renders, the `AlertModalContext.Provider` re-renders unnecessarily since `showAlert` and `clearAlert` actually doesn't need to be a new reference on every render. But with that said, it's a hypothesis, will need to try and measure the performance of the context changes

Gabrieltay: I tried to monitor to see if it re-renders unnecessarily. Didn't encounter it while doing some basic actions. I think we should monitor the perf first before introducing `useCallback` or `useMemo`. Maybe create a chore for perf testing/monitoring.
image

src/components/AlertModal/AlertModal.tsx
Gabrieltay: Initially, the reason I used the `useMemo` here, is to cater for users who stubbornly keep trying to execute the same action and facing the same error. But I realised that by clearing the alert modal, the `props` got overridden anyway, kinda defeat the purpose of the `useMemo`. So I will be removing the `useMemo` here. 
Based on the above [comments](https://github.com/rationally-app/mobile-application/pull/128#discussion_r465454367) , lets do a proper perf monitoring and implement `memorized` as needed.

src/components/AlertModal/AlertModal.tsx
Gabrieltay: nice! TIL.
Updated and pushed

src/components/AlertModal/AlertModal.tsx
```diff
@@ -0,0 +1,159 @@
...
+  alertType,
+  title,
+  description,
+  buttonTextType,
+  visible,
+  onOk,
+  onCancel,
+  onExit
+}) => {
+  const primaryButton = (): JSX.Element => {
```
sebastianquek, line 103: Minor nitpick, do we still need this as a function?

Gabrieltay: ok, changed to assignment.

PR #131, opened 2020-07-29
feat: add domain check

ilhaoni:
add domain check to verify valid backend


src/components/Login/LoginContainer.tsx
```diff
@@ -191,6 +191,9 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
       try {
         const { key, endpoint } = decodeQr(qrCode);
         Vibration.vibrate(50);
+        const endpointCheck = process.env.DOMAIN_CHECK as string;
```
sebastianquek, line 194: Another way to approach this is to set this as an `extra` field in `app.config.ts`, this provides the ability to check if it exists before continuing. So if someone were to try building the app without providing this prop, an error will show.

src/components/Login/LoginContainer.tsx
```diff
@@ -191,6 +191,9 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
...
+        if (!RegExp(endpointCheck).test(endpoint))
```
sebastianquek, line 195: How shall we handle this for development?

ilhaoni:
Ahh I see! Okk, will cater for the deeplink next!

src/components/Login/LoginContainer.tsx
ilhaoni: Good point, I'll make the change haha

src/components/Login/LoginContainer.tsx
ilhaoni: Hmm if we run and build it locally, we can actually configure our .env locally to accept the the PR endpoints. 

If we use the Github actions bot's channel, we'll have to provision for a domain for the PR haha. I saw a package previously that could add domain creation into serverless.yml. But I'm not sure if we should use the package though :/

What do you think?(:

ilhaoni: https://github.com/amplify-education/serverless-domain-manager

.github/workflows/deploy_prod.yml
```diff
@@ -71,6 +71,7 @@ jobs:
           SENTRY_DSN: ${{ secrets.SENTRY_DSN }}
           APP_BUILD_VERSION: ${{ github.run_number }}
           APP_BINARY_VERSION: ${{steps.latestBinaryVersion.outputs.version}}
```
sebastianquek, line 73: ohnu, i forgot to add the spaces, can help add here too :p

src/components/Login/LoginContainer.tsx
ilhaoni: Added!

src/components/Login/LoginContainer.tsx
ilhaoni: Actually I was thinking about it, I feel that the library feels too buggy to be used...

Should I just look into other ways of automating and posting the domain url? :/

.github/workflows/deploy_prod.yml
ilhaoni: done!

src/components/Login/LoginContainer.tsx
```diff
@@ -191,6 +195,8 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
       try {
         const { key, endpoint } = decodeQr(qrCode);
         Vibration.vibrate(50);
+        if (!RegExp(DOMAIN_CHECK).test(endpoint))
+          throw Error("Invalid endpoint");
```
sebastianquek, line 199: Small convention (:
```suggestion
          throw new Error("Invalid endpoint");
```

src/components/Login/LoginContainer.tsx
```diff
@@ -147,6 +148,9 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
   useEffect(() => {
     const skipScanningIfParamsInDeepLink = async (): Promise<void> => {
       const { queryParams } = await Linking.parseInitialURLAsync();
+      const queryEndpoint = queryParams?.endpoint;
+      if (queryEndpoint && !RegExp(DOMAIN_CHECK).test(queryEndpoint))
+        throw Error("Invalid endpoint");
```
sebastianquek, line 153:
```suggestion
        throw new Error("Invalid endpoint");
```

sebastianquek, line 153: Also realised this isnt triggering the error boundary, so we aren't catching the errors to send back to Sentry.

As discussed, we can use this approach: https://github.com/facebook/react/issues/14981#issuecomment-468460187

src/components/Login/LoginContainer.tsx
ilhaoni: made the change~

src/components/Login/LoginContainer.tsx
ilhaoni: made the change~

ilhaoni: I made some updates to capture the errors on Sentry haha, but I still face issue in triggering the ErrorBoundary haha, will need help on that 😬

sebastianquek: Hmm I tried it, works fine on my phone + simulator, if you see the red screen on dev, you got to press dismiss to see the error boundary. On prod the red screen won't show

src/components/Login/LoginContainer.tsx
```diff
@@ -144,17 +145,26 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
...
 
+  const setState = useState()[1];
   useEffect(() => {
-    const skipScanningIfParamsInDeepLink = async (): Promise<void> => {
+    const skipScanningIfParamsInDeepLink = async (): Promise<any> => {
       const { queryParams } = await Linking.parseInitialURLAsync();
+      const queryEndpoint = queryParams?.endpoint;
+      if (queryEndpoint && !RegExp(DOMAIN_CHECK).test(queryEndpoint)) {
+        const error = new Error(`Invalid endpoint: ${queryEndpoint}`);
+        Sentry.captureException(error);
```
sebastianquek, line 155: Don't need this since it'll be called in the error boundary.

src/components/Login/LoginContainer.tsx
```diff
@@ -191,11 +201,15 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
...
         Vibration.vibrate(50);
+        if (!RegExp(DOMAIN_CHECK).test(endpoint))
+          throw new Error("Invalid endpoint");
         setCodeKey(key);
         setEndpointTemp(endpoint);
         setIsLoading(false);
         setLoginStage("MOBILE_NUMBER");
       } catch (e) {
+        const error = new Error(`onBarCodeScanned ${e}`);
+        Sentry.captureException(error);
```
dalsontws, line 212: just to clarify! for this portion, after throwing error `Invalid endpoint`, is it intended for the user to just see `Invalid QR Code` as the error message?

src/components/Login/LoginContainer.tsx
```diff
@@ -144,17 +145,26 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
...
-    const skipScanningIfParamsInDeepLink = async (): Promise<void> => {
+    const skipScanningIfParamsInDeepLink = async (): Promise<any> => {
       const { queryParams } = await Linking.parseInitialURLAsync();
+      const queryEndpoint = queryParams?.endpoint;
+      if (queryEndpoint && !RegExp(DOMAIN_CHECK).test(queryEndpoint)) {
+        const error = new Error(`Invalid endpoint: ${queryEndpoint}`);
+        Sentry.captureException(error);
+        setState(() => {
+          throw error;
+        });
```
sebastianquek, line 158: Actually, come to think of it, maybe an alert will suffice, like in the normal case, what do you think?

package.json
```diff
@@ -65,6 +65,7 @@
     "babel-loader": "^8.0.6",
     "babel-plugin-inline-dotenv": "^1.6.0",
     "babel-preset-expo": "^8.2.3",
+    "dotenv": "^8.2.0",
```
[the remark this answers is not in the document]
ilhaoni: okk~

src/components/Login/LoginContainer.tsx
```diff
@@ -144,17 +145,26 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
     }
   }, [isLoading, endpoint, navigation, token, features]);
 
+  const setState = useState()[1];
   useEffect(() => {
-    const skipScanningIfParamsInDeepLink = async (): Promise<void> => {
+    const skipScanningIfParamsInDeepLink = async (): Promise<any> => {
```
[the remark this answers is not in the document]
ilhaoni: ah i forgot to change it back, was experimenting haha

src/components/Login/LoginContainer.tsx
ilhaoni: ok~

src/components/Login/LoginContainer.tsx
dalsontws: hmm, i wasn't able to see the error boundary too, its currently a sentry warning at the bottom of my screen, is that expected?

src/components/Login/LoginContainer.tsx
ilhaoni: hahaha, hmm it depends. 

If we do the error boundary, it may be better cause at least user can see that something went wrong haha.

But an alert works fine too~ Maybe we stick with this first? :O

sebastianquek: Yea am good with either!

src/components/Login/LoginContainer.tsx
ilhaoni: ahh I see what you mean! Previously the `Invalid QR Code` was there before this haha xD Maybe we can choose Invalid QR code as the only error thrown?

src/components/Login/LoginContainer.tsx
ilhaoni: ok~

package.json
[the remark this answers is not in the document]
ilhaoni: actually @sebastianquek, the tests are failing cause we removed it HAHA. I add it back first? 😂

src/components/Login/LoginContainer.tsx
```diff
@@ -147,14 +148,20 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
...
+        setLoginStage("SCAN");
+      } else if (queryParams?.key && queryParams?.endpoint) {
         setCodeKey(queryParams.key);
         setEndpointTemp(queryParams.endpoint);
         setLoginStage("MOBILE_NUMBER");
       }
     };
     skipScanningIfParamsInDeepLink();
-  }, []);
+  });
```
sebastianquek, line 164: This needs a dependency array, we only need this effect to run once

ilhaoni: added back!

PR #132, opened 2020-07-29
feat: display quantity for checkout if no identifierinput

ddwsd:
Displays quantity for checkout if identifierInput does not exist. This is mainly for the FFTH distribution. It will affect those distributions of type "REDEEM".

### DIFF
--- a/src/components/CustomerQuota/CheckoutSuccess/RedeemedItem.tsx
+++ b/src/components/CustomerQuota/CheckoutSuccess/RedeemedItem.tsx
@@ -12,26 +12,28 @@ export const RedeemedItem: FunctionComponent<{
   itemQuantities: ItemQuantities;
 }> = ({ itemQuantities }) => {
   const { getProduct } = useProductContext();
-  const { category, identifierInputs } = itemQuantities;
+  const { category, identifierInputs, quantities } = itemQuantities;
   const categoryName = getProduct(category)?.name ?? category;
 
   const identifierInputDisplay = identifierInputs
     ? getIdentifierInputDisplay(identifierInputs)
-    : "";
+    : null;
+
+  const quantitiesDisplay = Object.values(quantities)[0];
 
   return (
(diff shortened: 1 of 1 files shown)


src/components/CustomerQuota/CheckoutSuccess/RedeemedItem.tsx
```diff
@@ -12,26 +12,28 @@ export const RedeemedItem: FunctionComponent<{
...
+
+  const quantitiesDisplay = Object.values(quantities)[0];
 
   return (
     <View style={{ marginBottom: size(1.5) }}>
       <View style={sharedStyles.itemRow}>
         <AppText style={sharedStyles.itemHeaderText}>{categoryName}</AppText>
       </View>
-      {!!identifierInputDisplay && (
+      {
```
sebastianquek, line 29: Do we still need these parentheses?

ddwsd: oh yeah we dont oops

src/components/CustomerQuota/CheckoutSuccess/RedeemedItem.tsx
```diff
@@ -12,26 +12,28 @@ export const RedeemedItem: FunctionComponent<{
...
-  const { category, identifierInputs } = itemQuantities;
+  const { category, identifierInputs, quantities } = itemQuantities;
   const categoryName = getProduct(category)?.name ?? category;
 
   const identifierInputDisplay = identifierInputs
     ? getIdentifierInputDisplay(identifierInputs)
-    : "";
+    : null;
+
+  const quantitiesDisplay = Object.values(quantities)[0];
```
ilhaoni, line 22: Just wondering, will there be an edge case where quantities is null, or empty? Will this fail? :O

ddwsd: based on the types, no

PR #140, opened 2020-08-13
feat: update error boundary styles

sebastianquek:
* Refactored error boundary content into a new component so it's easier to showcase in Storybook
* Use new styling for the error boundary to match the alert dialogs

image


ilhaoni:
Just wondering, do you think the app text and error message should be displayed with a larger font ah? They're currently a bit small on my screen haha

Also, I missed the Paiseh HAHA :x

PR #145, opened 2020-08-19
feat: add use of new alert modals for checkout

dalsontws:


src/context/alert.tsx
```diff
@@ -1,10 +1,40 @@
...
+  FunctionComponent,
+  useState,
+  useCallback
+} from "react";
 import {
   AlertModal,
   AlertModalProps
 } from "../components/AlertModal/AlertModal";
 
+export const ERROR_MESSAGE = {
```
sebastianquek, line 13: I think all these errors have a period at the end too!

sebastianquek, line 13: Same as on the backend, will an enum be appropriate here?

src/context/alert.tsx
```diff
@@ -18,6 +48,64 @@ const defaultAlertProp: AlertModalProps = {
   onExit: () => {}
 };
 
+export const defaultWarningProp: AlertModalProps = {
```
sebastianquek, line 51: Maybe can use plural for these?

src/hooks/useCart/useCart.tsx
```diff
@@ -206,7 +207,7 @@ export const useCart = (
             )
           );
         } else {
-          setError(e);
+          setError(new Error(e.message));
```
sebastianquek, line 210: This will discard the type of error, what's the reason for this change?

src/hooks/useCart/useCart.tsx
```diff
@@ -323,8 +324,15 @@ export const useCart = (
...
+        // backend will throw general duplicate identifier message
+        if (
+          e.message === "Invalid Purchase Request: Duplicate identifier inputs"
+        ) {
+          setError(new Error(ERROR_MESSAGE.DUPLICATE_IDENTIFIER_INPUT));
+        } else {
+          setError(new Error(ERROR_MESSAGE.SERVER_ERROR));
+        }
         setCartState("DEFAULT");
-        setError(new Error("Couldn't checkout, please try again later"));
```
sebastianquek, line 327: I think we need the error to be thrown after setting the cart state:
https://github.com/rationally-app/mobile-application/pull/129/files

src/components/CustomerQuota/CustomerQuotaScreen.tsx
```diff
@@ -145,11 +144,74 @@ export const CustomerQuotaScreen: FunctionComponent<CustomerQuotaProps> = ({
...
+            description:
+              campaignFeatures?.campaignName === "TT Token"
+                ? ERROR_MESSAGE.INVALID_POD_INPUT
+                : ERROR_MESSAGE.INVALID_IDENTIFIER_INPUT,
+            onOk: () => clearError()
+          });
+          break;
+
+        default:
+          showAlert({
```
sebastianquek, line 207: Do you think a sentry error here will be useful so we know there's an unhandled error? Or should we just throw and let the error boundary take care of it?

src/hooks/useCart/useCart.tsx
```diff
@@ -249,7 +250,7 @@ export const useCart = (
   const updateCart: CartHook["updateCart"] = useCallback(
     (category, quantity, identifierInputs) => {
       if (quantity < 0) {
-        setError(new Error("Invalid quantity"));
+        setError(new Error(ERROR_MESSAGE.INVALID_QUANTITY));
```
sebastianquek, line 253: This works well for now. Am wondering if it's useful to throw a special `CartError` instead? Then we can scope the error messages to be under `CartError` i.e. `new CartError(INVALID_QUANTITY)`. Perhaps the alert props can be inside the error too.

sebastianquek (changes requested):

src/hooks/useCart/useCart.tsx
dalsontws: this is actually just setting the error, not throwing it. would this be fine?

src/context/alert.tsx
dalsontws: oh! i was referencing the previous error messages that didnt have a period behind! should we add that?

dalsontws: good idea! changed!

src/components/CustomerQuota/CustomerQuotaScreen.tsx
dalsontws: alright! will be throwing to error boundary!

src/context/alert.tsx
dalsontws: got it, added!

src/hooks/useCart/useCart.tsx
dalsontws: good idea for this! will add this as a task for the bigger PR, thanks!

src/context/alert.tsx
dalsontws: added periods after asking Eida!

src/hooks/useCart/useCart.tsx
dalsontws: reverting to use `setError(e)`, will be thrown to error boundary!

src/hooks/useCart/useCart.tsx
sebastianquek: Oh like it needs to be set after the cart state haha

dalsontws: okay got it and changed!

sebastianquek (approved):
LGTM!

PR #148, opened 2020-08-21
Feat:standardise login error msg

ddwsd:
Testing using [this PR](https://github.com/rationally-app/backend-api/pull/109)
PS: is there any reason why `services/auth/index.tsx` has the file extension `tsx` instead of `ts`
- [x] Update login error messages
- [x] Add different classes of login errors, each with their own default `alertModalProps`
- [x] Rename the errors messages thrown in `decodeQR` to be more relevant for us
- [x] Update error messages thrown in backend
- [x] Add a layer in `services/auth` to process the different types of errors received
- [x] Update `services/quota` to process errors regarding authorization
- [x] Force user logout if user receives authorization error when sending requests to endpoints requiring authentication in the backend

KIV
- Update `usePastTransactions`(?) (i looked at the code, seems like the case i am considering – able to fetch quota with current auth credentials but not past transactions immediately after – is highly unlikely)
- Use alert modals for the logout expiry alert (may end up with 2 modals at once if the user is at a screen showing modals already, would have to refactor `useValidateExpiry` as well)
- Clear alerts before showing logout alert modal in `useLogout`


src/components/Login/LoginOTPCard.tsx
```diff
@@ -120,20 +125,19 @@ export const LoginOTPCard: FunctionComponent<LoginOTPCard> = ({
...
-        showAlert({
-          ...disabledAccessAlertProps,
-          description: e.message,
-          onOk: () => resetStage()
-        });
+        showAlert({ ...e.alertProps, onOk: () => resetStage() });
+      } else if (e instanceof OTPWrongError) {
+        showAlert(e.alertProps);
+      } else if (e instanceof OTPExpiredError) {
+        showAlert(e.alertProps);
```
dalsontws, line 132: maybe can combine these 2 conditions?

ddwsd: sounds good!

src/components/Login/LoginOTPCard.tsx
```diff
@@ -126,6 +127,8 @@ export const LoginOTPCard: FunctionComponent<LoginOTPCard> = ({
         showAlert(e.alertProps);
       } else if (e instanceof LoginError) {
         showAlert(e.alertProps);
+      } else if (e instanceof AuthInvalidError) {
+        showAlert({ ...e.alertProps, onOk: () => resetStage() });
```
ddwsd: `AuthInvalidError` is an instance of `LoginError`, so 
```suggestion
      } else if (e instanceof AuthInvalidError) {
        showAlert({ ...e.alertProps, onOk: () => resetStage() });
      } else if (e instanceof LoginError) {
        showAlert(e.alertProps);
```

dalsontws: changed, thanks!!

src/components/Login/LoginOTPCard.tsx
```diff
@@ -120,20 +122,16 @@ export const LoginOTPCard: FunctionComponent<LoginOTPCard> = ({
...
+      } else if (e instanceof OTPWrongError || e instanceof OTPExpiredError) {
+        showAlert(e.alertProps);
+      } else if (e instanceof AuthInvalidError) {
+        showAlert({ ...e.alertProps, onOk: () => resetStage() });
       } else if (e instanceof LoginError) {
-        showAlert({
-          ...invalidInputAlertProps,
-          description: e.message
-        });
+        showAlert(e.alertProps);
```
dalsontws, line 131: actually thinking about this, for `LoginError`, whats the reason not to `resetStage`? in this case the user will be stuck at OTP screen. 
Also, maybe we can combine the types of errors in a single condition as well `(e instanceof LoginLockedError || e instanceof AuthInvalidError)

dalsontws (changes requested):

src/components/Login/LoginOTPCard.tsx
ddwsd: i think in this case the `LoginError` is usually because of network error. `resetStage` should only be because of errors in the qr code or when the user is locked out. what cases did u have in mind when you asked this question?

dalsontws: i was testing the use of a spoofed backend endpoint, the issue was the error message told me to try again, but in the end i couldn't because i was stuck in the otp card, need to kill the app.

ddwsd: hmm that makes sense too...spoofed backend -> throws network error -> stuck in otp card. ok will change it to `resetStage` as well haha

src/components/Login/LoginContainer.tsx
```diff
@@ -140,18 +143,12 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
...
+        showAlert({ ...e.alertProps, onOk: () => resetStage() });
+      } else if (e instanceof LoginLockedError) {
+        showAlert({ ...e.alertProps, onOk: () => resetStage() });
       } else if (e instanceof LoginError) {
-        showAlert({
-          ...invalidInputAlertProps,
-          description: e.message,
-          onOk: () => resetStage()
-        });
+        showAlert(e.alertProps);
```
dalsontws, line 151: might have to add `resetStage` here too!

dalsontws (changes requested):

src/components/Login/LoginContainer.tsx
dalsontws: it was stuck after spoofing the endpoint but providing a valid key.
and if we do add `resetStage` here, then we can combine all 3 error instances in this as one condition haha

ddwsd: yeap changed haha

src/services/envVersion/index.ts
```diff
@@ -29,6 +30,8 @@ const liveGetEnvVersion = async (
   } catch (e) {
     if (e instanceof ValidationError) {
       Sentry.captureException(e);
+    } else if (e.message === "Invalid authentication token provided") {
+      throw new AuthInvalidError(e.message);
     }
     throw new EnvVersionError(e.message);
```
dalsontws, line 36: hmm actually for this, any idea why `ValidationError` will just throw a `EnvVersionError`?
in this case, actually any type of error (including `ValidationError`) will be thrown as `EnvVersionError`, so maybe can remove the validation error condition, and also can capture all exceptions with sentry, wdyt?

dalsontws (changes requested):

src/services/auth/index.tsx
```diff
@@ -2,18 +2,111 @@ import { IS_MOCK } from "../../config";
...
+        ? ERROR_MESSAGE.LAST_OTP_ERROR
+        : ERROR_MESSAGE.OTP_ERROR
+    };
+  }
+}
+
+export class OTPExpiredError extends LoginError {
+  constructor(message: string) {
+    super(message);
+    this.name = "OTPWrongError";
```
sebastianquek, line 105:
```suggestion
    this.name = "OTPExpiredError";
```

src/services/quota/index.tsx
```diff
@@ -16,6 +16,13 @@ export class NotEligibleError extends Error {
   }
 }
 
+export class AuthenticationError extends Error {
```
sebastianquek, line 19: How is this different from the set of `Auth.*Error` in the auth service?

src/services/quota/index.tsx
```diff
@@ -142,6 +149,8 @@ export const liveGetQuota = async (
     }
     if (e.message === "User is not eligible") {
       throw new NotEligibleError(e.message);
+    } else if (e.message === "Invalid authentication token provided") {
+      throw new AuthenticationError(e.message);
```
sebastianquek, line 153: This error will occur whenever we try to use an invalid auth token on a protected endpoint, shall we move it to be inside the `fetchWithValidator` so we can ensure that for every protected endpoint, we could. get this auth error?

src/services/quota/index.tsx
ddwsd: this is not exactly a `LoginError`, so i didnt want to reuse the errors in the auth service. i named it differently to avoid confusion 😅

src/services/quota/index.tsx
ddwsd: sounds good!

src/services/helpers.ts
```diff
@@ -18,6 +25,12 @@ export async function fetchWithValidator<T, O, I>(
 
   const json = await response.json();
   if (!response.ok) {
+    // the first message is thrown by mystique, the sec
```
sebastianquek: I think you missed out on completing the sentence haha

src/services/envVersion/index.ts
dalsontws: will leave it as is!

src/services/quota/index.tsx
sebastianquek: It's challenging to differentiate between `AuthenticationError` and `AuthInvalidError`/`AuthExpiredError`. Can we name this better to avoid confusion?

ddwsd: how about `SessionError`?

sebastianquek: Sorry just saw this, Yep sounds good! Are there other types of session errors? Or only expiry will make sense?

src/hooks/useCart/useCart.tsx
```diff
@@ -203,16 +204,13 @@ export const useCart = (
...
           setCartState("NOT_ELIGIBLE");
-          // Cart will remain in FETCHING_QUOTA state.
+          return;
         } else if (e instanceof QuotaError) {
-          setError(
-            new Error(
-              "Error getting quota. We've noted this down and are looking into it!"
-            )
-          );
+          setError(new Error(ERROR_MESSAGE.QUOTA_ERROR));
```
sebastianquek, line 209: I noticed this discards the type of error (i.e. QuotaError), what if we let the consumer handle it instead? Is there an advantage of discarding QuotaError?

src/services/quota/index.tsx
ddwsd: np! we were mainly testing it using expiry (we manually change the expiry date in the db after logging in). but that shouldn't happen realistically because we have the `useValidateExpiry` hook and the backend returns a ttl on login. 
interestingly, our backend won't check for the role of the user in the entire login flow. so creating an user with role that is not "OPERATOR" would trigger this type of error also for any requests that go thru `onlyAuthorizedOperatorSession` in the backend

sebastianquek: I see, thanks for explaining! Ok sounds good then, for now we just stick to a generic SessionError, in future we may include other types.

src/hooks/useCart/useCart.tsx
ddwsd: sure!

sebastianquek (approved):
LGTM! Thanks for abstracting all the errors!!

dalsontws (approved):
LGTM!

PR #151, opened 2020-08-24
feat: finalise QR error messages

dalsontws:
Finalised QR error message
image

### DIFF
--- a/src/components/Login/utils.ts
+++ b/src/components/Login/utils.ts
@@ -22,9 +22,6 @@ export const decodeQr = (code: string): DecodedQrResponse => {
   } catch (e) {
     if (e.message.includes("Unexpected token"))
       throw new Error(ERROR_MESSAGE.AUTH_FAILURE_EXPIRED_TOKEN);
-    // if user is not an operator
-    if (e.message.includes("Unauthorized auth token"))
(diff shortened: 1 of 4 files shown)


src/components/Login/utils.ts
```diff
@@ -22,6 +22,9 @@ export const decodeQr = (code: string): DecodedQrResponse => {
   } catch (e) {
     if (e.message.includes("Unexpected token"))
       throw new Error(ERROR_MESSAGE.AUTH_FAILURE_EXPIRED_TOKEN);
+    // if user is not an operator
+    if (e.message.includes("Unauthorized auth token"))
+      throw new Error(ERROR_MESSAGE.AUTH_FAILURE_INVALID_TOKEN);
```
ddwsd, line 27: hmm in the decode QR stage, it is not possible to tell the role of the user, so i don't think this block should be here

ddwsd (commented):
i feel like this issue can be handled in #148 instead (the error message changes can be kept here tho) 😅

ddwsd (approved):
LGTM!

src/components/Login/utils.ts
dalsontws: got it, removed!

PR #152, opened 2020-08-25
feat: styling updates for alert with no description

sebastianquek:
image


ddwsd (approved):
LGTM! Do you think it's we should update storybook too?

sebastianquek:
@ddwsd Have added an example in storybook, thanks for the heads up!

PR #154, opened 2020-08-25
fix: change behavior of cancel in customer quota screen

yongjiajie:
[Notion Reference](https://www.notion.so/Cancel-button-to-go-back-to-Scan-ID-main-page-instead-of-camera-scanning-feature-30e737b54d1147ef82a6cef2c32f2e79)

[Related PR](https://github.com/rationally-app/mobile-application/pull/147)

This is a fix for bugs found in the related PR:

- [x]  After a user has selected an appeal reason and clicks back, the user will be redirected to the main page, instead of the Appeal page.
- [x]  Clicking Cancel at the Appeal page goes back to the previous page and not the main screen.


yongjiajie:
Closed this pull request as similar fixes have been done in another pull request. See #159.

PR #161, opened 2020-08-27
feat: show all past transactions on limit reached screen

szenius:
This PR adds 
- [x] Change `usePastTransactions` hook to call `/transactions/history` endpoint so it works for multiple IDs (prep for showing all past transactions on checkout success screen
- [x] Restyling of each transaction in limit reached screen
- [x] Display quantity with units for each transaction in limit reached screen
- [x] Group past transactions in limit reached screen by category
- [x] Order past transactions by latest transacted category, followed by the rest in order of category in policy
- [x] Show only first 5 transactions by default, hide the rest with "Show more"
- [x] Display appeal label for appeal transactions
- [x] Refactor `NoQuotaCard` into smaller files
- [x] Add/Update unit tests

### DIFF
--- a/.github/workflows/deploy_branch_preview.yml
+++ b/.github/workflows/deploy_branch_preview.yml
@@ -30,7 +30,7 @@ jobs:
       - name: Check Lint
         run: npm run lint
       - name: Set Timezone to +8 (Ubuntu)
-        run: sudo timedatectl set-timezone Asia/Shanghai
+        run: sudo timedatectl set-timezone Asia/Singapore
       - name: Test
         run: npm run test -- --coverage
   deploy_branch_preview:
--- a/.github/workflows/deploy_branch_preview.yml
+++ b/.github/workflows/deploy_branch_preview.yml
@@ -29,6 +29,8 @@ jobs:
         run: npx --no-install tsc --noEmit
       - name: Check Lint
         run: npm run lint
+      - name: Set Timezone to +8 (Ubuntu)
+        run: sudo timedatectl set-timezone Asia/Shanghai
       - name: Test
         run: npm run test -- --coverage
   deploy_branch_preview:
--- a/src/components/CustomerQuota/NoQuota/NoQuotaCard.test.ts
+++ b/src/components/CustomerQuota/NoQuota/NoQuotaCard.test.ts
@@ -1,6 +1,6 @@
 import { toDate } from "date-fns";
-import { defaultIdentifier } from "../../../test/helpers/wrapper";
-import { Policy, PastTransactionsResult } from "../../../types";
+import { defaultIdentifier } from "../../../test/helpers/defaults";
+import { PastTransactionsResult, CampaignPolicy } from "../../../types";
 import {
   groupTransactionsByCategory,
   sortTransactions,
@@ -9,7 +9,7 @@ import {
 
 describe("NoQuotaCard utility functions", () => {
   let sortedTransactions: PastTransactionsResult["pastTransactions"];
-  let allProducts: Policy[];
+  let allProducts: CampaignPolicy[];
   let mockTransactionsByCategoryMap: TransactionsByCategoryMap;
 
   const latestTransactionTimeMs = 1596530350000;
--- a/src/components/CustomerQuota/NoQuota/NoQuotaCard.test.ts
+++ b/src/components/CustomerQuota/NoQuota/NoQuotaCard.test.ts
@@ -1,0 +1,209 @@
+import { toDate } from "date-fns";
+import { defaultIdentifier } from "../../../test/helpers/wrapper";
+import { Policy, PastTransactionsResult } from "../../../types";
+import {
+  groupTransactionsByCategory,
+  sortTransactions,
+  TransactionsByCategoryMap
+} from "./NoQuotaCard";
+
+describe("NoQuotaCard utility functions", () => {
+  let sortedTransactions: PastTransactionsResult["pastTransactions"];
+  let allProducts: Policy[];
+  let mockTransactionsByCategoryMap: TransactionsByCategoryMap;
+
+  const latestTransactionTimeMs = 1596530350000;
+  const latestTransactionTime = toDate(latestTransactionTimeMs);
+
+  beforeAll(() => {
+    sortedTransactions = [
+      {
+        category: "tt-token",
+        quantity: 1,
+        identifierInputs: [
+          {
+            ...defaultIdentifier,
+            label: "Device code",
+            value: "AAA987654321"
+          }
+        ],
+        transactionTime: latestTransactionTime
+      },
+      {
+        category: "meal-credits",
+        quantity: 10,
+        transactionTime: latestTransactionTime
+      },
+      {
+        category: "vouchers",
+        quantity: 1,
+        identifierInputs: [
+          {
+            ...defaultIdentifier,
+            label: "Voucher code",
+            value: "AAA987654322"
+          }
+        ],
+        transactionTime: toDate(latestTransactionTimeMs - 10000)
+      },
+      {
+        category: "meal-credits",
+        quantity: 5,
+        transactionTime: toDate(latestTransactionTimeMs - 20000)
+      }
+    ];
+    allProducts = [
+      {
+        category: "tt-token",
+        name: "TT token",
+        order: 0,
+        quantity: { period: 1, limit: 1 },
+        identifiers: [
+          {
+            label: "Device code",
+            textInput: { disabled: true, visible: true, type: "STRING" },
+            scanButton: { disabled: false, visible: true, type: "QR" }
+          }
+        ]
+      },
+      {
+        category: "vouchers",
+        name: "CDC Vouchers",
+        order: 0,
+        quantity: {
+          period: 1,
+          limit: 1,
+          unit: { type: "POSTFIX", label: " book" }
+        },
+        identifiers: [
+          {
+            label: "Voucher code",
+            textInput: { disabled: false, visible: true, type: "STRING" },
+            scanButton: { disabled: false, visible: true, type: "BARCODE" }
+          }
+        ]
+      },
+      {
+        category: "meal-credits",
+        name: "Meal credits",
+        order: 0,
+        quantity: { period: 1, limit: 20, unit: { type: "PREFIX", label: "$" } }
+      }
+    ];
+
+    mockTransactionsByCategoryMap = {
+      "TT token": {
+        transactions: [
+          {
+            transactionDate: "4 Aug 2020, 4:39PM",
+            details: "AAA987654321",
+            quantity: "1 qty",
+            isAppeal: false
+          }
+        ],
+        hasLatestTransaction: true
+      },
+      "Meal credits": {
+        transactions: [
+          {
+            transactionDate: "4 Aug 2020, 4:39PM",
+            details: "",
+            quantity: "$10",
+            isAppeal: false
+          },
+          {
+            transactionDate: "4 Aug 2020, 4:38PM",
+            details: "",
+            quantity: "$5",
+            isAppeal: false
+          }
+        ],
+        hasLatestTransaction: true
+      },
+      "CDC Vouchers": {
+        transactions: [
+          {
+            transactionDate: "4 Aug 2020, 4:39PM",
+            details: "AAA987654322",
+            quantity: "1 book",
+            isAppeal: false
+          }
+        ],
+        hasLatestTransaction: false
+      }
+    };
+  });
+
+  describe("groupTransactionsByCategory", () => {
+    it("should return categories with latest transactions and other categories", () => {
+      expect.assertions(1);
+
+      expect(
+        groupTransactionsByCategory(
+          sortedTransactions,
+          allProducts,
+          latestTransactionTime
+        )
+      ).toStrictEqual(mockTransactionsByCategoryMap);
+    });
+
+    it("should return empty object if sortedTransactions is null", () => {
+      expect.assertions(1);
+
+      expect(
+        groupTransactionsByCategory(null, allProducts, undefined)
+      ).toStrictEqual({});
+    });
+  });
+
+  describe("sortTransactions", () => {
+    it("should return array of transactions by category, categories with latest transactions should be before the rest", () => {
+      expect.assertions(1);
+      expect(sortTransactions(mockTransactionsByCategoryMap)).toStrictEqual([
+        {
+          category: "Meal credits",
+          transactions: [
+            {
+              transactionDate: "4 Aug 2020, 4:39PM",
+              details: "",
+              quantity: "$10",
+              isAppeal: false,
+              order: 0
+            },
+            {
+              transactionDate: "4 Aug 2020, 4:38PM",
+              details: "",
+              quantity: "$5",
+              isAppeal: false,
+              order: 1
+            }
+          ]
+        },
+        {
+          category: "TT token",
+          transactions: [
+            {
+              transactionDate: "4 Aug 2020, 4:39PM",
+              details: "AAA987654321",
+              quantity: "1 qty",
+              isAppeal: false,
+              order: 2
+            }
+          ]
+        },
+        {
+          category: "CDC Vouchers",
+          transactions: [
+            {
+              transactionDate: "4 Aug 2020, 4:39PM",
+              details: "AAA987654322",
+              quantity: "1 book",
+              isAppeal: false,
+              order: 3
+            }
+          ]
+        }
+      ]);
+    });
+  });
+});
--- a/src/components/CustomerQuota/NoQuota/NoQuotaCard.test.ts
+++ b/src/components/CustomerQuota/NoQuota/NoQuotaCard.test.ts
@@ -1,4 +1,3 @@
-import { toDate } from "date-fns";
 import { defaultIdentifier } from "../../../test/helpers/defaults";
 import { PastTransactionsResult, CampaignPolicy } from "../../../types";
 import {
@@ -13,7 +12,7 @@ describe("NoQuotaCard utility functions", () => {
   let mockTransactionsByCategoryMap: TransactionsByCategoryMap;
 
   const latestTransactionTimeMs = 1596530350000;
-  const latestTransactionTime = toDate(latestTransactionTimeMs);
+  const latestTransactionTime = new Date(latestTransactionTimeMs);
 
   beforeAll(() => {
     sortedTransactions = [
@@ -44,12 +43,12 @@ describe("NoQuotaCard utility functions", () => {
             value: "AAA987654322"
           }
         ],
-        transactionTime: toDate(latestTransactionTimeMs - 10000)
+        transactionTime: new Date(latestTransactionTimeMs - 10000)
       },
       {
         category: "meal-credits",
         quantity: 5,
-        transactionTime: toDate(latestTransactionTimeMs - 20000)
+        transactionTime: new Date(latestTransactionTimeMs - 20000)
       }
     ];
     allProducts = [
--- a/src/components/CustomerQuota/NoQuota/NoQuotaCard.test.ts
+++ b/src/components/CustomerQuota/NoQuota/NoQuotaCard.test.ts
@@ -54,9 +54,15 @@ describe("NoQuotaCard utility functions", () => {
     ];
     allProducts = [
       {
+        category: "meal-credits",
+        name: "Meal credits",
+        order: 0,
+        quantity: { period: 1, limit: 20, unit: { type: "PREFIX", label: "$" } }
+      },
+      {
         category: "tt-token",
         name: "TT token",
-        order: 0,
+        order: 1,
         quantity: { period: 1, limit: 1 },
         identifiers: [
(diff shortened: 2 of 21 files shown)


szenius:
Randomly picked a few reviewers based on Github's suggestions, sorry that you were chosen 😆 🧙

src/components/CustomerQuota/NoQuota/NoQuotaCard.tsx
```diff
@@ -0,0 +1,210 @@
...
+    const { transactions, hasLatestTransaction } = value;
+    if (hasLatestTransaction) {
+      latestTransactionsByCategory.push({
+        category: key,
+        transactions: transactions
+      });
+    } else {
+      otherTransactionsByCategory.push({ category: key, transactions });
+    }
+  });
```
sebastianquek, line 115: Hmmm I'm not very clear what this part does together with the concatenation below. Is there are better way to phrase the purpose of these 2 blocks? HMW test these too?

src/components/CustomerQuota/NoQuota/NoQuotaCard.tsx
```diff
@@ -0,0 +1,210 @@
...
+              </AppText>
+              {transactionsByCategoryList.map(
+                (
+                  transactionsByCategory: TransactionsByCategory,
+                  index: number
+                ) => (
+                  <TransactionsByCategory
+                    key={index}
+                    maxTransactionsToDisplay={
+                      isShowFullList ? 99999 : MAX_TRANSACTIONS_TO_DISPLAY
```
sebastianquek, line 186: We probably need to handle pagination especially for recurring campaigns such as FFTH where the user could be using it for many months, any thoughts on how we should deal with this?

src/components/CustomerQuota/NoQuota/NoQuotaCard.tsx
```diff
@@ -0,0 +1,210 @@
...
+  const sortedTransactions = pastTransactionsResult?.pastTransactions;
+
+  const latestTransactionTime: Date | undefined =
+    sortedTransactions?.[0].transactionTime ?? undefined;
+  const now = new Date();
+  const secondsFromLatestTransaction = latestTransactionTime
+    ? differenceInSeconds(now, latestTransactionTime)
+    : -1;
+
+  const hasAppealProduct = (): boolean => {
```
sebastianquek, line 63: Minor thing, do we need this to be a function or will a boolean var suffice?

src/components/CustomerQuota/NoQuota/NoQuotaCard.tsx
```diff
@@ -0,0 +1,210 @@
...
+  onCancel,
+  onAppeal
+}) => {
+  const [isShowFullList, setIsShowFullList] = useState<boolean>(false);
+  const { getProduct, allProducts } = useProductContext();
+  const { token, endpoint } = useAuthenticationContext();
+
+  const policyType = cart.length > 0 && getProduct(cart[0].category)?.type;
+
+  const { pastTransactionsResult } = usePastTransaction(ids, token, endpoint);
```
sebastianquek, line 52: Not sure if this was discussed, do you think we should show some spinner while this is loading?

sebastianquek (commented):
Nice! Thanks for breaking down the NoQuotaCard into smaller components!! Have added some comments, mostly revolving around testability!

src/utils/getIdentifierInputDisplay.ts
```diff
@@ -12,6 +12,7 @@ const processIdentifierInputValue = (
     : identifierInput.value;
 };
 
+// TODO: @deprecated in favour of getAllIdentifierInputDisplay below
```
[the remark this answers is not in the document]
szenius: Yup!

src/components/CustomerQuota/NoQuota/NoQuotaCard.tsx
```diff
@@ -0,0 +1,210 @@
...
+  // Assumes results are already sorted (valid assumption for results from /transactions/history)
+  const sortedTransactions = pastTransactionsResult?.pastTransactions;
```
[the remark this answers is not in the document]
szenius: Yes, sorted by transaction time. Do you think adding a getter in `usePastTransactions` would be better? `getSortedTransactions`?

src/components/CustomerQuota/NoQuota/NoQuotaCard.tsx
```diff
@@ -0,0 +1,210 @@
...
+      otherTransactionsByCategory.push({ category: key, transactions });
+    }
+  });
+
+  // Order each category list by alphabetical order
+  const sortTransactionsByCategory = (
+    a: TransactionsByCategory,
+    b: TransactionsByCategory
+  ): 1 | -1 | 0 =>
+    a.category > b.category ? 1 : b.category > a.category ? -1 : 0;
```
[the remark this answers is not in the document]
szenius: It is being moved to util in the next PR for the redeemed screen. 

For category, Eida said to order like that:
* Category with latest transaction
* All other categories by alphabetical order

Confirmed with her, it's in the story too~

src/components/CustomerQuota/NoQuota/NoQuotaCard.tsx
```diff
@@ -0,0 +1,210 @@
...
+        item.quantity,
+        policy?.quantity.unit || { type: "POSTFIX", label: " qty" }
+      ),
+      isAppeal: policy?.categoryType === "APPEAL"
+    });
+    transactionsByCategoryMap[categoryName].hasLatestTransaction =
+      transactionsByCategoryMap[categoryName].hasLatestTransaction ||
+      differenceInSeconds(latestTransactionTime || 0, item.transactionTime) ===
+        0;
+  });
```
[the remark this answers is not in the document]
szenius: Hmmm ideally should be writing test for the functional component 🙈 Not sure about benefit of refactoring this because it is only used by this functional component. What do you think if I moved this to a method outside of the component and wrote a test for it? I'm thinking to keep the util function in the same file since only this file uses it

src/components/CustomerQuota/NoQuota/NoQuotaCard.tsx
szenius: Does my explanation about the requirements clear things up for you?

>For category, Eida said to order like that:
>
>Category with latest transaction
>All other categories by alphabetical order

Same suggestion about moving to util function but keep in same file?

src/components/CustomerQuota/NoQuota/NoQuotaCard.tsx
szenius: Does the policy refresh period last for months? Not sure what is the policy requirement for FFTH. The past transactions endpoint only returns transactions within the current policy period

src/components/CustomerQuota/NoQuota/NoQuotaCard.tsx
szenius: Ehhhhhhh it was already here so I can't argue for the original intention. I think what you said makes sense so I will change it~

src/components/CustomerQuota/NoQuota/NoQuotaCard.tsx
szenius: Not discussed, will let Eida know about this

src/components/CustomerQuota/NoQuota/NoQuotaCard.tsx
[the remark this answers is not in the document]
sebastianquek: Yep, agreed, lets move it outside the component, then write tests for it. Alternatively can write integration tests for this noQuota screen 😊

src/components/CustomerQuota/NoQuota/NoQuotaCard.tsx
sebastianquek: Oh right ok, thought there was still that usecase to show all. Why do we need the `99999` here?

src/components/CustomerQuota/NoQuota/NoQuotaCard.tsx
sebastianquek: heheh thanks!

src/components/CustomerQuota/NoQuota/NoQuotaCard.tsx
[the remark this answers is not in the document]
sebastianquek: Was initially thinking just return the `pastTransactions` directly, so something like:
```
setPastTransactionResult(result?.pastTransactions)
```

src/components/CustomerQuota/NoQuota/NoQuotaCard.tsx
sebastianquek: Yea would be best if we can add some tests here (:

src/components/CustomerQuota/NoQuota/NoQuotaCard.tsx
szenius: Eh, the "show more"? We show first 5 transactions by default, and show everything within policy refresh period when user clicks "Show more"

src/components/CustomerQuota/NoQuota/NoQuotaCard.tsx
szenius: Made the change~

src/components/CustomerQuota/NoQuota/NoQuotaCard.tsx
szenius: From discussion with Eida, this will be covered in another story~

src/components/CustomerQuota/NoQuota/NoQuotaCard.tsx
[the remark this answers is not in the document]
szenius: Done~

src/components/CustomerQuota/NoQuota/NoQuotaCard.tsx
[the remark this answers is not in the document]
szenius: Moved it outside of the function component for now, next story will move to utils file~

src/components/CustomerQuota/NoQuota/NoQuotaCard.tsx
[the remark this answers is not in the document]
szenius: Done!! Broke them up in a way that I think makes sense, do let me know if you think can be broken down more :)

src/components/CustomerQuota/NoQuota/NoQuotaCard.tsx
szenius: Done!! Broke them up in a way that I think makes sense, do let me know if you think can be broken down more :)

src/hooks/usePastTransaction/usePastTransaction.tsx
```diff
@@ -9,32 +9,32 @@ export type PastTransactionHook = {
...
       );
 
       setPastTransactionsResult(pastTransactionsResponse);
     };
 
-    if (prevId != id)
+    if (prevIds !== ids)
       fetchPastTransactions().catch(() => {
-        Sentry.captureException(`Unable to fetch past transactions: ${id}`);
+        Sentry.captureException(`Unable to fetch past transactions: ${ids}`);
```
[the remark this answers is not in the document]
szenius: Added displaying alert dialogue when an error is thrown!! I went back and forth on how to structure the hook to throw errors, would appreciate if you have any inputs :)

szenius:
@sebastianquek Hmm looks like the tests are failing because we are formatting the epoch time in UTC +0, but locally it's passing for me (formatted in +8). It also used to pass in the CI (was formatted to +8). I am not familiar with how the timezone is being set (environment variables?) Did something change?

src/components/CustomerQuota/NoQuota/NoQuotaCard.test.ts
```diff
@@ -0,0 +1,215 @@
...
+            isAppeal: false
+          }
+        ],
+        hasLatestTransaction: true,
+        order: 1
+      },
+      "Meal credits": {
+        transactions: [
+          {
+            transactionDate: "4 Aug 2020, 4:39PM",
```
szenius, line 110: @sebastianquek Hmm looks like the tests are failing because we are formatting the epoch time in UTC +0, but locally it's passing for me (formatted in +8). It also used to pass in the CI (was formatted to +8). I am not familiar with how the timezone is being set (environment variables?) Did something change?

Also I checked in the app by running a local server, it also displays +8 leh

szenius: Spoke with seb, it's a locale issue -- on local it is +8 and on CI it is +0. Working on a fix now for the CI locale to be configured properly

szenius: Fixed :D

src/components/CustomerQuota/NoQuota/NoQuotaCard.tsx
```diff
@@ -0,0 +1,296 @@
...
+      quantity: formatQuantityText(
+        item.quantity,
+        policy?.quantity.unit || { type: "POSTFIX", label: " qty" }
+      ),
+      isAppeal: policy?.categoryType === "APPEAL"
+    });
+    transactionsByCategoryMap[categoryName].hasLatestTransaction =
+      transactionsByCategoryMap[categoryName].hasLatestTransaction ||
+      differenceInSeconds(latestTransactionTime || 0, item.transactionTime) ===
+        0;
```
sebastianquek, line 99: Would checking if `latestTransactionTime === item.transactionTime` suffice?

src/hooks/usePastTransaction/usePastTransaction.test.tsx
```diff
@@ -128,83 +124,87 @@ describe("usePastTransaction", () => {
...
+          category: "tt-token",
+          quantity: 1,
+          identifierInputs: [
+            {
+              ...defaultIdentifier,
+              label: "first",
+              value: "AAA987654321"
+            }
+          ],
+          transactionTime: toDate(1596530356430)
```
sebastianquek, line 147: Noticed several uses of `toDate`, any idea what's the advantage of this over `new Date(...)`?

src/hooks/usePastTransaction/usePastTransaction.tsx
```diff
@@ -3,40 +3,49 @@ import { PastTransactionsResult } from "../../types";
...
     };
 
-    if (prevId != id)
-      fetchPastTransactions().catch(() => {
-        Sentry.captureException(`Unable to fetch past transactions: ${id}`);
-      });
-  }, [authKey, endpoint, id, prevId]);
+    if (
+      typeof prevIds !== typeof ids ||
+      (prevIds && ids && !prevIds.every(id => ids.includes(id)))
```
sebastianquek, line 41: Hmm does `prevIds !== ids` work?
Also, the 2nd condition implies that if `ids` has an additional id appended to `prevIds`, then `fetchPastTransactions` won't be called, when we want it to be called

src/components/CustomerQuota/NoQuota/NoQuotaCard.tsx
sebastianquek: Would it be cleaner if `maxTransactionsToDisplay` is optional, then we can avoid this big number?

src/hooks/usePastTransaction/usePastTransaction.tsx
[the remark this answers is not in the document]
sebastianquek: Yep looks good! I think we should clear the error whenever `fetchPastTransactions` is called too, so we don't have some stale error state lingering on

src/hooks/usePastTransaction/usePastTransaction.tsx
szenius: No eh, cuz `[1] !== [1]`. We were doing the fetch multiple times with just `prevIds !== ids`.

You're right, let me do a equal length check too

src/components/CustomerQuota/NoQuota/NoQuotaCard.tsx
szenius: No, because in the backend we add a millisecond to each item in the same transaction. So need to round to seconds to check instead~

src/hooks/usePastTransaction/usePastTransaction.test.tsx
szenius: I am not sure, copied this from other parts of the codebase 🙈  I guess it is consistent with our usage of `date-fns` everywhere?

src/components/CustomerQuota/NoQuota/NoQuotaCard.tsx
szenius: Hmm I did consider this but that will make for more complex conditional display logic in `TransactionsByCategory`. It would look like

```
export const TransactionsByCategory: FunctionComponent<{
  category: string;
  transactions: Transaction[];
  maxTransactionsToDisplay: number | undefined;
}> = ({ category, transactions, maxTransactionsToDisplay }) => {
  const shouldShowCategory =
    maxTransactionsToDisplay === undefined ||
    (transactions[0].order || 0) < maxTransactionsToDisplay;
  return shouldShowCategory ? (
    <View style={styles.wrapper}>
      <View style={styles.itemRow}>
        <AppText style={styles.itemHeader}>{category}</AppText>
      </View>
      {transactions.map(
        (transaction, index) =>
          (maxTransactionsToDisplay === undefined ||
            (transaction.order || 0) < maxTransactionsToDisplay) && (
            <Transaction key={index} {...transaction} />
          )
      )}
    </View>
  ) : null;
};

```

Seems to me it is quite prone to mistake. Of course we can do testing for it but am not sure if it might be cleaner just to pass in the big number. 

I previously declared this as a constant instead. Does that make it better (since no more magic number in the middle of code)?

src/hooks/usePastTransaction/usePastTransaction.tsx
sebastianquek: In this case the passed in `ids` won't change reference since it comes from the ids state in `CustomerQuotaScreen`. So when the effect runs, the fetch won't be called unless a new id was added.

I think checking each value is useful, but doesn't work too well if we're comparing arrays of objects. I'm wondering what's a good pattern/guideline for these sort of checks as we use this elsewhere too (e.g. useCart).

src/hooks/usePastTransaction/usePastTransaction.test.tsx
sebastianquek: Yea i saw it then was like uhoh, can we stick to `new Date(...)` instead? 😊 
Can use `date-fns` only when necessary perhaps

.github/workflows/deploy_branch_preview.yml
```diff
@@ -29,6 +29,8 @@ jobs:
         run: npx --no-install tsc --noEmit
       - name: Check Lint
         run: npm run lint
+      - name: Set Timezone to +8 (Ubuntu)
+        run: sudo timedatectl set-timezone Asia/Shanghai
```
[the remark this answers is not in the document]
sebastianquek: yay looks like it works

src/components/CustomerQuota/NoQuota/NoQuotaCard.tsx
sebastianquek: Yep ok, thanks for explaining the tradeoffs!

.github/workflows/deploy_branch_preview.yml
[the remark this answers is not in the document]
szenius: Good point, made the change~

src/hooks/usePastTransaction/usePastTransaction.tsx
szenius: Hmm that's weird but i logged it and observed that even though the ids stayed the same, the fetch kept happening. I am not so sure about the data path here but let me test it again.

src/hooks/usePastTransaction/usePastTransaction.test.tsx
sebastianquek: thanks!! (:

src/hooks/usePastTransaction/usePastTransaction.tsx
szenius: Turns out it was the unit test that was set up wrongly. Original `prevIds !== ids` check would have sufficed for the app itself. Reverted :)

src/hooks/usePastTransaction/usePastTransaction.test.tsx
szenius: Changed!

src/hooks/usePastTransaction/usePastTransaction.tsx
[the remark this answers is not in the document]
szenius: Good point!! Done~

src/hooks/usePastTransaction/usePastTransaction.tsx
sebastianquek: Oh saw the offending test, a good reminder too haha

sebastianquek (approved):
Super awesome work!!! 💯 🏅

PR #163, opened 2020-08-28
fix: bug fix for checkoutLimit

justin0108:
Bug: when checkout limit is 1, upon toggle of itemCheckBox on checkout, the quantity will be change to maxQuantity


sebastianquek (approved):
Thanks for fixing this!!

PR #165, opened 2020-08-28
chore: useQuota hook

millasml:
https://www.notion.so/701364d4fbbf4ec1801b766c589973b9?v=24aa9675206d4da9a171ca4073d81b5b&p=2c332462475940f0884703d9ba4fb8a4

Creating a separate useQuota hook out of the useCart hook

- [x] create useQuota hook
- [x] update useCart to remove all quota related things (as much as possible)
- [x] fix tests for useCart
- [x] move some test to useQuota
- [x] update codebase to use useQuota

**Potentially Breaking Changes**
- `@testing-library/react-hooks` is updated from `3.2.1` to `3.4.2`, which changes how `rerender` works.


src/hooks/useQuota/useQuota.tsx
```diff
@@ -0,0 +1,114 @@
...
+};
+
+const hasNoQuota = (quota: Quota): boolean =>
+  quota.remainingQuota.every(item => item.quantity === 0);
+
+const hasInvalidQuota = (quota: Quota): boolean =>
+  // Note: Invalid quota refers to negative quota received
+  quota.remainingQuota.some(item => item.quantity < 0);
+
+export const useQuota = (
```
sebastianquek, line 47: Do add tests for this!

src/hooks/useQuota/useQuota.tsx
```diff
@@ -0,0 +1,114 @@
...
+import { Quota, Policy } from "../../types";
+
+export type QuotaHook = {
+  quotaResponse: Quota | null;
+  allQuotaResponse: Quota | null;
+  fetchQuota: (setter: (cartState: CartState) => void) => Promise<void>;
+  quotaError?: Error;
+};
+
+type CartState =
```
sebastianquek, line 17: What's the reason for letting `useQuota` know about the `CartState`? How might we structure this hook so that it is agnostic on where it's called, and stays true to its single responsibility?

src/hooks/useCart/useCart.tsx
```diff
@@ -165,55 +136,28 @@ export const useCart = (
...
+  );
   const { showAlert } = useContext(AlertModalContext);
   const clearError = useCallback((): void => setError(undefined), []);
 
+  useEffect(() => {
+    if (quotaError) {
+      setError(quotaError);
+    }
+  }, [quotaError]);
+
```
sebastianquek, line 152: Haven't looked closely into the code, what is the reason for mirroring this state?

sebastianquek (commented):
Did a quick first pass!

src/hooks/useQuota/useQuota.tsx
```diff
@@ -0,0 +1,151 @@
...
+          setQuotaError(
+            new Error(
+              "Error getting quota. We've noted this down and are looking into it!"
+            )
+          );
+        } else {
+          Sentry.addBreadcrumb({
+            category: "useQuota",
+            message: "fetchQuota - unidentified error"
+          });
```
yongjiajie, line 130: Do we still need this breadcrumb here?

src/hooks/useQuota/useQuota.tsx
```diff
@@ -0,0 +1,151 @@
...
+        );
+        setQuotaResponse(filteredQuotas);
+      } catch (e) {
+        if (e instanceof NotEligibleError) {
+          setQuotaState("NOT_ELIGIBLE");
+        } else if (e instanceof QuotaError) {
+          Sentry.addBreadcrumb({
+            category: "useQuota",
+            message: "fetchQuota - quota error"
+          });
```
yongjiajie, line 120: Do we still need this breadcrumb here?

src/components/CustomerQuota/CustomerQuotaScreen.tsx
```diff
@@ -166,22 +174,22 @@ export const CustomerQuotaScreen: FunctionComponent<CustomerQuotaProps> = ({
   ]);
```
yongjiajie, line 174: Perhaps we can add another breadcrumb here to track the `quotaState` like how it's done with `cartState`? Not sure if it will add value, what do you think?

src/components/CustomerQuota/CustomerQuotaScreen.tsx
```diff
@@ -256,10 +264,22 @@ export const CustomerQuotaScreen: FunctionComponent<CustomerQuotaProps> = ({
...
-  return cartState === "FETCHING_QUOTA" ? (
+  useEffect(() => {
+    /**
+     * Update quota after checkout
+     */
+    if (cartState === "PURCHASED" && !updateAfterPurchased) {
+      updateQuota();
+      setUpdateAfterPurchased(true);
+    }
+  }, [cartState, updateQuota, updateAfterPurchased]);
```
yongjiajie, line 280: Looks good, we only need to update the quota once after purchase after all 👍

src/hooks/useCart/useCart.tsx
```diff
@@ -131,114 +115,19 @@ const mergeWithCart = (
...
-const hasInvalidQuota = (quota: Quota): boolean =>
-  // Note: Invalid quota refers to negative quota received
-  quota.remainingQuota.some(item => item.quantity < 0);
-
 export const useCart = (
   ids: string[],
   authKey: string,
-  endpoint: string
+  endpoint: string,
+  quotaResponse?: Quota | null
```
yongjiajie, line 122: @sebastianquek please advise if this is okay 😅

yongjiajie (dismissed):

src/components/CustomerQuota/CustomerQuotaScreen.tsx
millasml: yeah! i realized it gets stuck in an update loop so this was necessary to get out of it!

src/hooks/useCart/useCart.tsx
sebastianquek: Perhaps we can think of what the cart needs to know? I noticed `Quota` now contains different types of quota, does the cart need to know about all the different types?

Also, when will this be `null` and not `undefined`?

millasml: quotaResponse might be null because in the useQuota hook, we initialize the quotaResponse state to be null. Upon rendering, the useQuota hook should fetch the quota and replace the quotaResponse with an actual value. However, in the event that an error occurs, the quotaResponse will still remain null.

src/components/CustomerQuota/CustomerQuotaScreen.tsx
```diff
@@ -192,6 +204,7 @@ export const CustomerQuotaScreen: FunctionComponent<CustomerQuotaProps> = ({
           break;
 
         case ERROR_MESSAGE.MISSING_IDENTIFIER_INPUT:
+          console.log("show error");
```
yongjiajie, line 207: Remember to remove all your `console.log`s.

src/hooks/useQuota/useQuota.tsx
```diff
@@ -0,0 +1,143 @@
...
+};
+
+export const useQuota = (
+  ids: string[],
+  authKey: string,
+  endpoint: string
+): QuotaHook => {
+  const [quotaState, setQuotaState] = useState<QuotaState>("DEFAULT");
+  const [quotaResponse, setQuotaResponse] = useState<Quota | null>(null);
+  const [allQuotaResponse, setAllQuotaResponse] = useState<Quota | null>(null);
```
sebastianquek, line 84: What's the reason for using `null` rather than doing `useState<Quota>()`?

src/hooks/useQuota/useQuota.tsx
```diff
@@ -0,0 +1,143 @@
...
+        setQuotaResponse(filteredQuotas);
+      } catch (e) {
+        if (e instanceof NotEligibleError) {
+          setQuotaState("NOT_ELIGIBLE");
+          return;
+        } else {
+          Sentry.addBreadcrumb({
+            category: "useQuota",
+            message: "fetchQuota - unidentified error"
+          });
```
sebastianquek, line 121: What's the purpose of this breadcrumb?

src/hooks/useQuota/useQuota.tsx
```diff
@@ -0,0 +1,143 @@
...
+          setQuotaState("NO_QUOTA");
+        } else {
+          setQuotaState("DEFAULT");
+        }
+        setAllQuotaResponse(quota);
+        const filteredQuotas = filterQuotaWithAvailableProducts(
+          quota,
+          products
+        );
+        setQuotaResponse(filteredQuotas);
```
sebastianquek, line 112: What's the difference between all quota vs quota? Can we comment in the `QuotaHook` type?

src/hooks/useQuota/useQuota.tsx
```diff
@@ -0,0 +1,143 @@
...
+      if (products.some(policy => policy.category === itemQuota.category))
+        result.remainingQuota.push(itemQuota);
+    },
+    filteredQuota
+  );
+
+  if (quota.globalQuota) {
+    filteredQuota.globalQuota = [];
+    transform(
+      quota.globalQuota!,
```
sebastianquek, line 38: Noticed we're asserting the existence of this, why is this necessary?

src/hooks/useQuota/useQuota.tsx
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  return filteredQuota;
+};
+
+const hasNoQuota = (quota: Quota): boolean => {
+  if (quota === null) {
+    return true;
+  }
+  return quota.remainingQuota.every(item => item.quantity === 0);
```
sebastianquek, line 66:
```suggestion
  return quota === null || quota.remainingQuota.every(item => item.quantity === 0);
```

Actually how come we need to check for `null`? The type of the argument is `Quota` and not `Quota | null`, when will it be `null`?

src/hooks/useQuota/useQuota.tsx
```diff
@@ -0,0 +1,143 @@
...
+      }
+    };
+    update();
+  }, [ids, authKey, endpoint, products, quotaResponse]);
+
+  useEffect(() => {
+    if (prevIds !== ids || prevProducts !== products) {
+      updateQuota();
+    }
+  }, [prevIds, ids, prevProducts, products, updateQuota]);
```
sebastianquek, line 134: I noticed this hook automatically calls to get quota and also exposes the function so it can be called imperatively. I think we should add it in the comments of the `useQuota` hook so the developer who  uses it can know that it does both of these.

src/hooks/useQuota/useQuota.test.tsx
```diff
@@ -0,0 +1,291 @@
...
+      const { result } = renderHook(() => useQuota(ids, key, endpoint), {
+        wrapper
+      });
+
+      expect(result.current.quotaState).toBe("NOT_ELIGIBLE");
+    });
+  });
+
+  describe("update quota when necessary", () => {
+    it("should update quota ids are added", async () => {
```
sebastianquek, line 248:
```suggestion
    it("should update quota when ids are added", async () => {
```

src/hooks/useQuota/useQuota.test.tsx
```diff
@@ -0,0 +1,291 @@
...
+
+      const { result } = renderHook(() => useQuota(ids, key, endpoint), {
+        wrapper
+      });
+
+      expect(result.current.quotaState).toBe("NOT_ELIGIBLE");
+    });
+  });
+
+  describe("update quota when necessary", () => {
```
sebastianquek, line 247: There's also another test when the `products` from the context changes

src/components/CustomerQuota/CustomerQuotaScreen.tsx
```diff
@@ -256,10 +267,31 @@ export const CustomerQuotaScreen: FunctionComponent<CustomerQuotaProps> = ({
...
+  }, [cartState, updateQuota, updateAfterPurchased]);
+
+  useEffect(() => {
+    /**
+     * Initialise available cart items when quota or product changes
+     */
+    if (quotaResponse?.remainingQuota) {
+      initCartWithQuota(quotaResponse.remainingQuota);
+    }
+  }, [quotaResponse, initCartWithQuota, products]);
```
sebastianquek, line 292: What are your thoughts if we pass it as a prop into `useCart` instead? Something like `initialCartQuota`, then inside the hook, an `useEffect` merges the `initialCartQuota` with the existing cart whenever it changes.

sebastianquek (changes requested):

src/hooks/useQuota/useQuota.tsx
yongjiajie: I have changed this to:
```suggestion
  const [quotaResponse, setQuotaResponse] = useState<Quota>();
  const [allQuotaResponse, setAllQuotaResponse] = useState<Quota>();
```

src/hooks/useQuota/useQuota.tsx
yongjiajie: This has been removed!

src/hooks/useQuota/useQuota.tsx
yongjiajie: This was done because `quota.globalQuota` was declared as a `partial` type. I think we should move forward with declaring both `quota.globalQuota` and `quota.localQuota` as non-partial, since they are always returned.

Let me know what you think.

src/hooks/useQuota/useQuota.tsx
yongjiajie: I have added comments for this, let me know if they are sufficient.

src/hooks/useQuota/useQuota.test.tsx
yongjiajie: This has been added.

src/components/CustomerQuota/CustomerQuotaScreen.tsx
yongjiajie: I agree with this, this will inform developers that there is a dependency on `useQuota` as well.
This has been implemented.

src/hooks/useQuota/useQuota.test.tsx
yongjiajie: This is done!

src/hooks/useQuota/useQuota.tsx
```diff
@@ -0,0 +1,142 @@
...
+    quota.remainingQuota,
+    (result: Quota, itemQuota) => {
+      if (products.some(policy => policy.category === itemQuota.category))
+        result.remainingQuota.push(itemQuota);
+    },
+    filteredQuota
+  );
+
+  transform(
+    quota.globalQuota!,
```
sebastianquek, line 41: Noticed the assertion, guess we don't need this anymore? EDIT: noticed there are a couple of assertions elsewhere too, perhaps can be removed too

package.json
```diff
@@ -53,7 +53,7 @@
     "@storybook/addons": "^5.3.9",
     "@storybook/react-native": "^5.3.21",
     "@testing-library/jest-native": "^3.1.0",
-    "@testing-library/react-hooks": "3.2.1",
+    "@testing-library/react-hooks": "3.4.2",
```
sebastianquek, line 56: We'll needa verify that existing tests that use `rerender` do not depend on the new prop argument incorrectly too

src/components/CustomerQuota/CheckoutSuccess/CheckoutSuccessCard.tsx
```diff
@@ -40,7 +40,7 @@ const styles = StyleSheet.create({
 interface CheckoutSuccessCard {
   ids: string[];
   onCancel: () => void;
-  quotaResponse: Quota | null;
+  quotaResponse?: Quota;
```
sebastianquek, line 43: Hmmm minor distinction, by having it as `?` we imply that the quotaResponse can be excluded, which isn't the case. Having it as `quotaResponse: Quota | undefined` implies instead that we *must* pass in a `quotaResponse` just that it could possibly be undefined. This prevents future errors when `CheckoutSuccessCard` is used but no `quotaResponse` was passed in.

src/hooks/useCart/useCart.test.tsx
```diff
@@ -182,152 +213,20 @@ describe("useCart", () => {
...
-        wrapper
-      });
-
-      expect(result.current.cartState).toBe("NOT_ELIGIBLE");
-    });
-  });
-
   describe("update cart quantities", () => {
-    it("should update the cart when more ids are added", async () => {
+    it("should update the cart when quota response changes", () => {
```
sebastianquek, line 217: Hmm this test case doesn't seem to be testing when quota response changes, noticed it's just the ids that are changing?

src/hooks/useCart/useCart.test.tsx
```diff
@@ -182,152 +213,20 @@ describe("useCart", () => {
...
+        () =>
+          useCart(ids, key, endpoint, mockQuotaResMultipleIds.remainingQuota),
         { wrapper }
       );
 
-      await waitForNextUpdate();
-
-      mockGetQuota.mockReturnValueOnce(mockQuotaResMultipleIds);
       ids = ["ID1", "ID2"];
-      rerender([ids, key, endpoint]);
```
sebastianquek, line 328: Right, this is actually the point on upgrading react hooks testing lib, we can't do this anymore. Looks like it really was the useCart tests lolol @yongjiajie

src/hooks/useCart/useCart.test.tsx
```diff
@@ -466,15 +313,17 @@ describe("useCart", () => {
...
-      });
+      const { result } = renderHook(
+        () => useCart(ids, key, endpoint, mockQuotaResSingleId.remainingQuota),
+        {
+          wrapper
+        }
+      );
 
-      await wait(() => {
+      act(() => {
```
sebastianquek, line 326: thanks for changing these!!

src/hooks/useCart/useCart.tsx
```diff
@@ -131,148 +111,48 @@ const mergeWithCart = (
...
-  /**
-   * Merge quota response with current cart whenever quota response or products change.
+   * Update the cart when:
+   *  1. First quota is retrieved on initialisation
+   *  2. When quota is updated (i.e. products or ids change)
    */
   useEffect(() => {
-    if (quotaResponse) {
-      // Note that we must use a callback within this setState to avoid
-      // having cart as a dependency which causes an infinite loop.
```
sebastianquek, line 248: I think we should leave this in for future devs to take note of the potential infinite loop

src/hooks/useCart/useCart.test.tsx
```diff
@@ -182,152 +213,20 @@ describe("useCart", () => {
...
-          category: "chocolate",
-          descriptionAlert: undefined,
-          identifierInputs: [],
-          lastTransactionTime: transactionTime,
-          maxQuantity: 0,
-          quantity: 0
-        }
-      ]);
-    });
-    it("should have cart state be NO_QUOTA when quota received is invalid", async () => {
```
sebastianquek, line 264: I noticed we removed this test case, and also in useQuota hook we could potentially return negative values for quantities, can we make sure the cart will always receive valid (i.e. non-negative quantities) values?

src/hooks/useQuota/useQuota.test.tsx
```diff
@@ -0,0 +1,488 @@
...
+      mockGetQuota.mockImplementation(() => {
+        mockIdNotEligible(ids[0]);
+      });
+
+      const { result } = renderHook(() => useQuota(ids, key, endpoint), {
+        wrapper
+      });
+
+      expect(result.current.quotaState).toBe("NOT_ELIGIBLE");
+    });
```
sebastianquek, line 409: Hmmm not sure what this test is doing, what does it mean by would not continue with fetching quota? HMW make this clearer?

src/types.ts
```diff
@@ -173,12 +173,11 @@ const ItemQuota = t.intersection([
...
-  t.type({
-    remainingQuota: t.array(ItemQuota)
-  }),
-  t.partial({ localQuota: t.array(ItemQuota), globalQuota: t.array(ItemQuota) })
-]);
+export const Quota = t.type({
+  remainingQuota: t.array(ItemQuota),
+  localQuota: t.array(ItemQuota),
+  globalQuota: t.array(ItemQuota)
+});
```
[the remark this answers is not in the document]
yongjiajie: I have went through and made sure `localQuota` and `globalQuota` are not checked for `undefined`!

package.json
yongjiajie: I have refactored all `rerender` to not include `props`.

Moving forward, we can simply reassign the `props` used by the hook, then call `rerender` to call the hook with the new `props`.

```typescript
let ids = ["ID1"];
const { rerender, result } = renderHook(
  () => useCart(ids, ...),
  { wrapper }
)

// Say we want to update ids in useCart hook
ids = ["ID1", "ID2"];
rerender();
```

The `props` in `rerender` will now be used to update the `wrapper` context values.
```typescript
let ids = ["ID1"];
const { rerender, result } = renderHook(
  () => useCart(ids, ...),
  { wrapper }
)

// Say we want to update the context in wrapper
rerender({ products: newProducts });
```

src/components/CustomerQuota/CheckoutSuccess/CheckoutSuccessCard.tsx
yongjiajie: I agree with this. This is done!

src/hooks/useCart/useCart.test.tsx
yongjiajie: You're welcome!

src/hooks/useCart/useCart.tsx
yongjiajie: I agree, this has been returned to the respective location!

src/hooks/useCart/useCart.test.tsx
yongjiajie: Our back-end actually does not return negative values since there are checks in place, so this test is redundant. Hence, I'm not sure if there's a possibility of the cart receiving negative quantities.

We might have to revisit if there is even a possibility of an invalid quota existing (i.e. negative quantity).

sebastianquek: yea granted the backend right now does not return invalid values, an accidental mishap could mean the frontend showing strange states.

This is also why we used io-ts, to make sure the backend data isnt malformed to a certain extent. The additional layer of checking of negative values isn't covered by that validation right after the fetch though.

src/hooks/useCart/useCart.test.tsx
yongjiajie: I have changed this to make sure that we change the quota response, and test that it changes accordingly after `rerender`.

src/hooks/useCart/useCart.test.tsx
yongjiajie: I have added another check in `useCart` such that if `cartQuota` contains negative quantities, an error is thrown, and the cart is not set.

src/hooks/useQuota/useQuota.test.tsx
yongjiajie: I have renamed this to: `"should set quota state to NOT_ELIGIBLE when NotEligibleError is thrown, and does not update quota response"`.

I have also added another test case to test that if an existing quota response exists and `NotEligibleError` is thrown, the quota state is `NOT_ELIGIBLE`, but the previous quota response is not overridden. Let me know if these tests are sound.

src/hooks/useQuota/useQuota.tsx
yongjiajie: This no longer exists in the code, let me know if this is still an issue.

src/hooks/useCart/useCart.tsx
yongjiajie: This no longer exists in the code, let me know if this is still an issue.

src/hooks/useCart/useCart.tsx
yongjiajie: This no longer exists in the code, let me know if this is still an issue.

src/hooks/useQuota/useQuota.tsx
yongjiajie: After discussion, I have moved forward with declaring `quota.globalQuota`  and `quota.localQuota` as non-partial.

ilhaoni:
hi hi~ since this is approved, let's aim to merge this in by this week? (: Great work!

sebastianquek (approved):

yongjiajie:
A bug was found where the appeal flow was affected.

`hasNoQuota()` in the `useQuota` hook will have to be refactored to take into account the new quota response, as well as the appeal flow.

PR #166, opened 2020-08-28
feat: use new campaign config across the app

sebastianquek:
**Warning: large PR, I will walk through the changes with the reviewers** 😅 

Here's a rough summary of the changes:
* Remove use of the `envVersion` service
* Updated `CampaignInitialisationScreen` to handle all the retrieval/checking of config, previously was in `LoginContainer`
* New `CampaignConfigsStoreContext` that holds all the campaign configs across different campaigns and interfaces with async storage
 * Refactored `useUpdateCampaignConfig` to do the hashing when the `updateCampaignConfig` is called, rather than store the generated hashes for all campaigns in memory
* Updated `CampaignConfigContext` provides a way to scope the campaign config for all descendants in a particular tree 
* New `AuthStoreContext` that holds all auth credentials and interfaces with async storage
* Updated `AuthContext` provides a way to scope the auth credentials for all descendants in a particular tree
  * Auth credentials now include `operatorToken`, `sessionToken`, `endpoint` and `expiry`
  * Have added a temporary migration path to move old credentials to the new `map` based way so that we can hold multiple auth credentials in app for different campaigns
  * Note that the keys to these `map`s are `${operatorToken}${endpoint}`
* Updated `CustomerQuotaStack` to dynamically provide the `AuthContext` and `CampaignConfigContext` to all screens in this stack
* Updated all the quota components to use the appropriate `ProductContext`
* Fixed issue with phone number input not using the correct font family
* Removed some unnecessary errors in teh validation utils

Todo:
- [x] Make sure auth errors are handled properly when fetching campaign config
- [x] merchant flow

How to test
* All changes should not cause any changes to current flow
* Going from an old version to this, the user should remain logged in


src/components/CampaignInitialisation/CampaignInitialisationScreen.tsx
```diff
@@ -49,38 +56,87 @@ export const CampaignInitialisationScreen: FunctionComponent<NavigationProps> =
...
+    hasLoadedFromStore,
+    updateCampaignConfig
+  ]);
 
   useEffect(() => {
-    if (fetchCampaignConfigError) {
-      throw fetchCampaignConfigError; // Let ErrorBoundary handle
+    if (updateCampaignConfigError) {
+      if (updateCampaignConfigError instanceof CampaignConfigError) {
+        Sentry.captureException(updateCampaignConfigError);
```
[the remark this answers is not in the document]
sebastianquek: Oh in this case it'll be thrown to the error boundary to handle the sending to Sentry

src/context/campaignConfig.tsx
```diff
@@ -1,90 +1,20 @@
...
+import React, { createContext, FunctionComponent } from "react";
+import { CampaignFeatures, CampaignConfig, CampaignPolicy } from "../types";
 
 interface CampaignConfigContext {
-  features: CampaignFeatures | null;
-  configHashes: ConfigHashes;
-  setCampaignConfig: (config: CampaignConfig) => void;
-  clearCampaignConfig: () => void;
+  readonly features: CampaignFeatures | null;
+  readonly policies: CampaignPolicy[] | null;
```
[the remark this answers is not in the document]
sebastianquek: Hmmm because it's an `array of CampaignPolicy`, but it assigns to a `policies` variable 😆

src/hooks/useCart/useCart.tsx
```diff
@@ -158,20 +158,14 @@ export const useCart = (
   endpoint: string
 ): CartHook => {
   const prevIds = usePrevious(ids);
-  const {
-    products,
-    getProduct,
-    setProducts,
-    setFeatures
-  } = useProductContext();
+  const { products, getProduct } = useContext(ProductContext);
```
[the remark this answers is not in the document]
sebastianquek: It gets 1 product though, not the entire array

src/navigation/CustomerQuotaStack/index.tsx
```diff
@@ -27,4 +33,33 @@ const CustomerQuotaStack = createStackNavigator(
...
+  const key = `${operatorToken}${endpoint}`;
+
+  const { authCredentials } = useContext(AuthStoreContext);
+  if (!authCredentials[key]) {
+    throw new Error("No auth credentials found");
+  }
+
+  const { allCampaignConfigs } = useContext(CampaignConfigsStoreContext);
+  if (!allCampaignConfigs[key]) {
+    throw new Error("No campaign config found");
```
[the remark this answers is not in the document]
sebastianquek: Good catch, i've updated it in the other PR, let me bring the changes back here instead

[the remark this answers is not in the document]
sebastianquek: Actually i realised the other PR handles it differently and removes the need for this error checking: https://github.com/rationally-app/mobile-application/pull/168/files#diff-ce64e55e6f9ca212a5eaff2d2d39ff0f

src/services/campaignConfig/campaignConfig.test.ts
```diff
@@ -14,24 +15,97 @@ const mockValidResponse = {
...
+          label: "$"
+        }
+      }
+    }
+  ]
 };
 
 const mockValidResponseNewFeature = {
   features: {
+    newFeature: true,
```
ilhaoni, line 61: is this just a dummy value to show that there's a change in the response HAHA

src/utils/validateIdentification.ts
```diff
@@ -1,26 +1,22 @@
...
     case "NRIC":
-      if (idRegex)
-        throw new EnvVersionError(
-          "We are currently facing connectivity issues. Try again later or contact your in-charge if the problem persists."
-        );
       id = validateAndCleanNric(inputId);
       break;
     case "REGEX":
+      if (!idRegex) {
+        throw new CampaignConfigError(ERROR_MESSAGE.ENV_VERSION_ERROR);
```
ilhaoni, line 18: nit: haha should we rename `ENV_VERSION_ERROR` to `CAMPAIGN_CONFIG_ERROR`?

src/context/campaignConfig.tsx
ilhaoni, line 6: ooh, til, having it as readonly is good~ ^^

src/context/campaignConfigsStore.tsx
```diff
@@ -0,0 +1,147 @@
...
+        return {
+          ...prevConfigs,
+          [key]: mergedConfig
+        };
+      });
+    },
+    []
+  );
+
+  const removeCampaignConfig: CampaignConfigsStoreContext["removeCampaignConfig"] = useCallback(
```
ilhaoni, line 90: hmm just curious. Given that we use `AsyncStorage.getItem`, should we use `AsyncStorage.removeItem` when we have to remove a particular campaign config as well? :O

src/context/campaignConfigsStore.tsx
```diff
@@ -0,0 +1,147 @@
...
+      setAllConfigs(prevConfigs => {
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        const { [key]: configToBeRemoved, ...remainingConfigs } = prevConfigs;
+        return remainingConfigs;
+      });
+    },
+    []
+  );
+
+  const clearCampaignConfigs: CampaignConfigsStoreContext["clearCampaignConfigs"] = useCallback(async () => {
```
ilhaoni, line 104: similar for this haha, i notice we dont use `AsyncStorage.removeItem` :O

src/context/campaignConfigsStore.tsx
sebastianquek: Hmmm we store all the configs under 1 key though, so it isn't really a `removeItem`. Do you think we should split into different keys in async storage?

src/components/Login/LoginContainer.tsx
```diff
@@ -133,12 +124,17 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
...
+  const setState = useState()[1];
   const handleRequestOTP = async (
     fullNumber = mobileNumber
   ): Promise<boolean> => {
     try {
       if (!(await getResendConfirmationIfNeeded())) return false;
-      const response = await requestOTP(fullNumber, codeKey, endpointTemp);
+      const response = await requestOTP(
+        fullNumber,
+        tempAuthCredentials?.operatorToken ?? "",
```
yongjiajie, line 135: May be a noob question, but why is it possible for `operatorToken` and `endpoint` to be empty? Is this for dev environment?

src/components/Login/LoginContainer.tsx
```diff
@@ -133,12 +124,17 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
...
+        tempAuthCredentials?.endpoint ?? ""
```
yongjiajie, line 136: May be a noob question, but why is it possible for operatorToken and endpoint to be empty? Is this for dev environment?

src/context/campaignConfigsStore.tsx
ilhaoni: hmm, I think it might be better, i was thinking there's 2 possible benefits:
- if there's any case of malformed data, we only affect one campaign config, instead of everything 
- in the event of a campaign-leak, at least only a single campaign is leaked haha

what do you think? ^^

the cons is definitely harder to query though. 🤔

src/context/authStore.test.tsx
```diff
@@ -0,0 +1,284 @@
...
+    const button = getByText("test button");
+    fireEvent.press(button);
+    await wait(() => {
+      expect(mockSetItem).toHaveBeenCalledTimes(1);
+      expect(mockSetItem).toHaveBeenCalledWith("AUTH_STORE", "{}");
+      expect(queryByTestId("credentials")).toHaveTextContent("");
+    });
+  });
+
+  it("should add a set of auth credentials properly", async () => {
```
ilhaoni, line 168: for this test, we're testing the case where its a new set of auth credentials being added. Do you think we should have a case to show that we append a new set of auth credentials to an existing set?

I'm assuming authStore works the same way as campaignConfigStore haha, so all the auths under 1 key, unless I'm wrong haha

src/context/authStore.tsx
```diff
@@ -0,0 +1,182 @@
...
+export const AuthStoreContext = createContext<AuthStoreContext>({
+  hasLoadedFromStore: false,
+  authCredentials: {},
+  addAuthCredentials: () => undefined,
+  removeAuthCredentials: () => undefined,
+  clearAuthCredentials: () => undefined
+});
+
+export const AuthStoreContextProvider: FunctionComponent<{
+  shouldMigrate?: boolean; // Temporary toggle to disable migration when testing
```
ilhaoni, line 45: so we'll set this to true in stg and prod? :O haha

src/components/CampaignInitialisation/utils.ts
```diff
@@ -0,0 +1,53 @@
...
+  const currentBinaryVersionSplit = currentBinaryVersion.split(".").map(Number);
+  const minBinaryVersionSplit = minBinaryVersion.split(".").map(Number);
+  for (let i = 0; i < minBinaryVersionSplit.length; i++) {
+    if (currentBinaryVersionSplit[i] < minBinaryVersionSplit[i]) {
+      return false;
+    } else if (currentBinaryVersionSplit[i] > minBinaryVersionSplit[i]) {
+      return true;
+    }
+  }
+  return true;
```
ilhaoni, line 22: hmm, so this will trigger when currentBinaryVersionSplit[i] = minBinaryVersionSplit[i]? 😆  

it's interesting to have a fallback `true` value for this function haha

src/context/campaignConfigsStore.tsx
sebastianquek: Yea, means that it needs to be async every time, would prefer to put it in "RAM" first, so querying is quick. Also, deletions are easy if we want to flush what's stored on async storage

Regarding the pt on malformed data, we can only control on the backend, unless we want  to validate on frontend too, which I think is not worth the effort. If the user decides to tweak it on his own, he'll reach invalid states and be brought to the error boundary.

Not sure what you mean by campaign leak, we don't encrypt anything haha, so if you compromise async storage, you'll see everything already regardless of key

src/context/authStore.tsx
sebastianquek: It's always true, just that for testing it's set to false.

src/components/Login/LoginContainer.tsx
sebastianquek: TypeScript will complain that `requestOTP` requires non-undefined params if we don't set the default to empty string. If instead we update the signature of `requestOTP` to take in undefined variables, then we're indicating that `requestOTP` has optional params, which should not be the case.

If you're wondering why `tempAuthCredentials` can be undefined, it's because we don't initialise it to any particular default value. The call to `handleRequestOTP` will only be called after `operatorToken` and `endpoint` have been defined.

How do you think this can be better approached?

src/components/CampaignInitialisation/utils.ts
sebastianquek: Yep when all values are the same.

What do you think will be a better approach for this?

src/services/campaignConfig/campaignConfig.test.ts
sebastianquek: yea, to indicate that if there's a new property  not defined in the current `policies` or `features`,  nothing breaks, it'll be safely ignored

ilhaoni: ahh, okok! ^^

src/context/campaignConfigsStore.tsx
ilhaoni: i see i see, that makes sense. hmmm, you are right haha, I blur xD 

i agree with the point on malformed data haha, lets do it the current way then! ^^

src/components/CampaignInitialisation/utils.ts
ilhaoni: oooh, hmm, what if we do

```
else if (currentBinaryVersionSplit[i] >= minBinaryVersionSplit[i])
```

sebastianquek: Hmm that won't work for
```
currentBinaryVersion = "3.1.0"
minBinaryVersionSplit = "3.2.0"
```
since once we reach the first index (i.e. 3), it'll match an return true

src/context/authStore.test.tsx
sebastianquek: Have just added the tests!

src/components/CampaignInitialisation/utils.ts
ilhaoni: hmm that is true. hmmm.

ilhaoni: I cant think of anything cleaner than this HAHA.

i thought of doing 

if current index 0 value === min index 0 value, check if current index 1 value > min index 1 value
else
check and compare from current index 1 and min index 1

but i think your way is cleaner

src/components/Login/LoginContainer.tsx
yongjiajie: I think it's fine! I was just curious as to why this was the case. If the call is only made after they have been defined then it is fine.

src/components/CampaignInitialisation/utils.ts
yongjiajie: Why do we have to split it? Can we just `join` instead of `split` and compare the raw numbers? For example if `currentBinaryVersion` is `3.1.0` and `minBinaryVersionSplit` is `3.2.0` it'll just be comparing `310` and `320`?

ilhaoni: :O that may work! so converting it into a number, and then compare~

src/navigation/CustomerQuotaStack/index.tsx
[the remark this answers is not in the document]
dalsontws: oh! so how is it handled if `allCampaignConfigs[key]` is undefined?

ilhaoni (approved):
LGTM! :D Thanks for the great work~ :D:D 

For me the remaining is the discussion on how to check the current and min version/binary, but the rest I've tested and it's good to go :D

src/components/CampaignInitialisation/utils.ts
sebastianquek: Because it can be 3.2.123, not limited to 0-9

ilhaoni: ah.... so if its current is 3.2.123 and min is 3.3.0, converting to numbers will not work... ok then, let's keep it this way first HAHA

yongjiajie: If that is the case, `split` makes sense. Perhaps the idea of `currentBinarySplit[n] - minBinaryVersionSplit[n]` can be used so the fallback is to return `false`?

So if `currentBinarySplit[n] - minBinaryVersionSplit[n] >= 0` return `true`, else return `false`, with a fallback of return `false`?

yongjiajie (approved):
All looks good to me!

src/components/CampaignInitialisation/utils.ts
sebastianquek: @yongjiajie yea that has the issue where it doesnt work for this:
```
currentBinaryVersion = "3.1.0"
minBinaryVersionSplit = "3.2.0"
expected to fail, but will pass because 3 - 3 = 0  => true
```

src/navigation/CustomerQuotaStack/index.tsx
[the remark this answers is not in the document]
sebastianquek: It'll navigate back to the list of campaign locations. Will only push to prod once these 2 PRs are merged!

src/components/CampaignInitialisation/utils.ts
yongjiajie: I think it's best to go ahead with your approach, looks like it covers the cases it needs to, and is simple!

PR #168, opened 2020-09-01
feat: multiple campaigns

sebastianquek:
TODO
- [x] Handle expired CampaignLocationListItem
- ~~Handle session expiry properly => should remove the auth credentials rather than remove all credentials~~ Will remove this feature for now

# ~~** REMOVE DELETE BUTTON BEFORE MERGE **~~

### DIFF
--- a/src/components/CampaignLocations/CampaignLocationsScreen.tsx
+++ b/src/components/CampaignLocations/CampaignLocationsScreen.tsx
@@ -6,7 +6,7 @@ import React, {
   useCallback
 } from "react";
 import { View, StyleSheet } from "react-native";
-import { size } from "../../common/styles";
+import { size, fontSize } from "../../common/styles";
 import { Credits } from "../Credits";
 import { KeyboardAvoidingScrollView } from "../Layout/KeyboardAvoidingScrollView";
 import { TopBackground } from "../Layout/TopBackground";
@@ -24,6 +24,7 @@ import { CampaignConfigsStoreContext } from "../../context/campaignConfigsStore"
 import { useDrawerContext } from "../../context/drawer";
 import { LoadingView } from "../Loading";
 import { Card } from "../Layout/Card";
+import { AppText } from "../Layout/AppText";
 
 const styles = StyleSheet.create({
   content: {
@@ -40,7 +41,18 @@ const styles = StyleSheet.create({
   bannerWrapper: {
     marginBottom: size(1.5)
   },
-  loadingViewWrapper: { alignItems: "center", justifyContent: "center" }
+  loadingViewWrapper: {
+    alignItems: "center",
+    justifyContent: "center"
+  },
+  selectCampaignHeader: {
+    fontFamily: "brand-bold",
+    fontSize: fontSize(1)
+  },
+  campaignLocationWrapper: {
+    marginHorizontal: -size(3),
+    marginTop: size(2)
+  }
 });
 
 export const CampaignLocationsScreen: FunctionComponent<NavigationProps> = ({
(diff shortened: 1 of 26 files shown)


src/components/CampaignLocations/CampaignLocationsScreen.tsx
```diff
@@ -0,0 +1,181 @@
...
+      }
+    ];
+
+    drawerButtons.push({
+      icon: "map-search",
+      label: "Change campaign",
+      onPress: () => {
+        navigation.navigate("CampaignLocationsScreen");
+      }
+    });
```
ilhaoni, line 99: Hmm, actually since we're showing this all the time, do we wanna add it to the array `drawerButtons` above?

src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
```diff
@@ -168,11 +213,14 @@ const CollectCustomerDetailsScreen: FunctionComponent<NavigationFocusInjectedPro
...
-                {features.campaignName}
-              </AppText>
-            )}
+            <View style={{ flexDirection: "row", alignItems: "flex-start" }}>
+              {features?.campaignName && (
+                <AppText style={styles.campaignName}>
+                  {features.campaignName}
+                </AppText>
+              )}
+              <ManageButton onPress={onPressDelete} />
```
ilhaoni, line 222: I think we wanna delete this right haha

src/components/CustomerQuota/CustomerQuotaScreen.tsx
```diff
@@ -235,13 +241,16 @@ export const CustomerQuotaScreen: FunctionComponent<CustomerQuotaProps> = ({
       throw new Error(error.message);
     }
   }, [
+    setAuthCredentials,
+    campaignFeatures?.campaignName,
```
ilhaoni, line 245: hmm, just curious, do we need this?

src/components/Login/LoginContainer.tsx
```diff
@@ -149,13 +171,14 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
     }
   };
```
ilhaoni, line 172: small nit, for this function, do you think we should rename `fullNumber` to `fullMobileNumber`? for consistency lol. if too small a nit, can ignore me also 😂

src/components/CustomerQuota/CustomerQuotaScreen.tsx
```diff
@@ -155,7 +148,20 @@ export const CustomerQuotaScreen: FunctionComponent<CustomerQuotaProps> = ({
...
+        sessionToken,
+        expiry: new Date().getTime()
+      });
+      showAlert({
+        ...expiredAlertProps,
+        description: ERROR_MESSAGE.AUTH_FAILURE_INVALID_TOKEN,
+        onOk: () => {
+          navigation.navigate("CampaignLocationsScreen");
+        }
+      });
```
ilhaoni, line 164: can extract this to be just `expireSession()` like one of the files below haha

src/context/config.tsx
```diff
@@ -15,6 +15,10 @@ export enum AppMode {
 
 interface Config {
   appMode: AppMode;
+  mobileNumber?: {
```
ilhaoni, line 18: maybe we can name this as `fullMobileNumber`? haha

ilhaoni (changes requested):
reviewed~ have some comments and questions hehe

src/components/CustomerQuota/CustomerQuotaScreen.tsx
sebastianquek: Yep, this was automatically added because of the check on the campaign name in the effect

ilhaoni: oooh, okok! i see!

ilhaoni:
hmm, I experienced a bug when I first loaded the app with a previous login session on Expo, the first screen I saw was "session expired", and then it brought me to the campaign initialisation screen. 

but when i add back the same campaign, it works fine :O is it an issue with migration?

ilhaoni (approved):
LGTM :D:D Thanks for the awesome work~~

PR #169, opened 2020-09-02
Feat/react nav 5 upgrade

dhemutton:
This PR contains the deprecation of react-navigation v4 modules and an upgrade to @react-navigation v5 with integration with navigation.


package.json
```diff
@@ -15,12 +15,17 @@
     "lint:fix": "npm run lint -- --fix"
   },
   "dependencies": {
+    "@react-native-community/masked-view": "^0.1.10",
```
sebastianquek, line 18: This dependency should be removed, it's not used plus the docs mentions "Note: React Native MaskedView is not currently supported by Expo unless you "eject"."

dhemutton: oh just checked, in https://reactnavigation.org/docs/getting-started, @react-native-community/masked-view is a dependency cos of the safe area view

sebastianquek:
Unfortunately this PR is too stale, we'll relook React Nav 5 another time.

PR #172, opened 2020-09-08
chore: Standardize CTA button text

yongjiajie:
[Notion Reference](https://www.notion.so/Ensure-that-the-Next-Identity-button-is-standardized-across-distributions-0d77eeafada646f9bbd933f0a10aac60)

The Call To Action buttons on some cards were inconsistent, with some showing "Next identity" and others showing "Next customer" based on the product type (i.e. `REDEEM` or `PURCHASE`). We want to standardize this across all distributions - it is now changed to "Next identity".


src/components/CustomerQuota/CheckoutSuccess/checkoutMessages.ts
```diff
@@ -17,7 +17,7 @@ export const getCheckoutMessages = (productType?: string): CheckoutMessage => {
       return {
         title: "Purchased!",
         description: "The following have been purchased:",
-        ctaButtonText: "Next customer"
+        ctaButtonText: "Next identity"
```
sebastianquek, line 20: Wondering if we should take out this property instead, since both product types result with the same text, what do you think?

sebastianquek:
We actually wanted to deprecate this specific "REDEEM" and "PURCHASE" prop. Instead, we can use the new campaign configuration on the param store,  `c13n/english`. https://www.notion.so/Campaign-Configuration-48cf81d1504d407cb860539e87fe26dc#5f1470a302fa42169a9bdd7dc8b86934

This will play in nicely with the translation story we have in the pipeline. Perhaps this PR can lay the groundwork to using this.

sebastianquek (approved):
LGTM!!! thanks for making this change!

yongjiajie:
Thanks @millasml for helping resolve the merge conflicts!

PR #175, opened 2020-09-11
feat: Display global limits messages only if there is a single category

yongjiajie:
This is related to #170 and #171 .

Global limits messages should only display if there is a single item category for now, since we have not decided how to show global limits messages for policies with multiple categories.

### Tests

- [x] `ffth-belanja-stg` should show global limits messages
- [x] `ffth-cfp-stg` should **not** show global limits messages
- [x] Every other distribution should **not** show global limits messages

### DIFF
--- a/src/components/CustomerQuota/CheckoutSuccess/CheckoutSuccessCard.tsx
+++ b/src/components/CustomerQuota/CheckoutSuccess/CheckoutSuccessCard.tsx
@@ -171,10 +171,10 @@ export const CheckoutSuccessCard: FunctionComponent<CheckoutSuccessCard> = ({
      * since we have not catered for showing global limit messages for multiple
      * categories.
      */
-    quotaResponse.globalQuota.length === 1 &&
+    allProducts?.length === 1 &&
     !!getProduct(sortedTransactions[0].category)?.quantity.usage;
 
-  const globalQuota = showGlobalQuota
+  const firstGlobalQuota = showGlobalQuota
     ? quotaResponse!.globalQuota![0]
     : undefined;
 
@@ -190,12 +190,10 @@ export const CheckoutSuccessCard: FunctionComponent<CheckoutSuccessCard> = ({
             />
             <AppText style={sharedStyles.statusTitleWrapper}>
               <AppText style={sharedStyles.statusTitle}>{title}</AppText>
-              {showGlobalQuota &&
-              globalQuota!.quantity &&
-              globalQuota!.quotaRefreshTime ? (
+              {showGlobalQuota && firstGlobalQuota!.quotaRefreshTime ? (
                 <UsageQuotaTitle
-                  quantity={globalQuota!.quantity}
-                  quotaRefreshTime={globalQuota!.quotaRefreshTime}
+                  quantity={firstGlobalQuota!.quantity}
+                  quotaRefreshTime={firstGlobalQuota!.quotaRefreshTime}
                 />
               ) : undefined}
             </AppText>
--- a/src/components/CustomerQuota/CheckoutSuccess/CheckoutSuccessCard.tsx
+++ b/src/components/CustomerQuota/CheckoutSuccess/CheckoutSuccessCard.tsx
@@ -172,7 +172,7 @@ export const CheckoutSuccessCard: FunctionComponent<CheckoutSuccessCard> = ({
      * categories.
      */
     allProducts?.length === 1 &&
-    !!getProduct(sortedTransactions[0].category)?.quantity.usage;
+    !!allProducts[0].quantity.usage;
 
   const firstGlobalQuota = showGlobalQuota
     ? quotaResponse!.globalQuota![0]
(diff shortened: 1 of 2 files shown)


src/components/CustomerQuota/CheckoutSuccess/CheckoutSuccessCard.tsx
```diff
@@ -181,16 +191,13 @@ export const CheckoutSuccessCard: FunctionComponent<CheckoutSuccessCard> = ({
...
-                  ({ quantity, quotaRefreshTime }, index: number) =>
-                    quotaRefreshTime ? (
-                      <UsageQuotaTitle
-                        key={index}
-                        quantity={quantity}
-                        quotaRefreshTime={quotaRefreshTime}
-                      />
-                    ) : undefined
-                )}
+              globalQuota!.quantity &&
```
szenius, line 194: Does this mean that if the quantity is 0, we don't want to show the title as well?

src/components/CustomerQuota/CheckoutSuccess/CheckoutSuccessCard.tsx
```diff
@@ -166,8 +166,18 @@ export const CheckoutSuccessCard: FunctionComponent<CheckoutSuccessCard> = ({
...
     sortedTransactions.length > 0 &&
+    /**
+     * We only display global limit messages if there is only one global quota,
+     * since we have not catered for showing global limit messages for multiple
+     * categories.
+     */
+    quotaResponse.globalQuota.length === 1 &&
     !!getProduct(sortedTransactions[0].category)?.quantity.usage;
 
+  const globalQuota = showGlobalQuota
```
szenius, line 177: Hmmm thinking how we might be able to improve the naming to reflect that this is different from the list of global quota items `quoteResponse.globalQuota`. `firstGlobalQuota`?

src/components/CustomerQuota/CheckoutSuccess/CheckoutSuccessCard.tsx
```diff
@@ -166,8 +166,18 @@ export const CheckoutSuccessCard: FunctionComponent<CheckoutSuccessCard> = ({
     !!quotaResponse?.globalQuota &&
     !!sortedTransactions &&
     sortedTransactions.length > 0 &&
+    /**
+     * We only display global limit messages if there is only one global quota,
+     * since we have not catered for showing global limit messages for multiple
+     * categories.
+     */
+    quotaResponse.globalQuota.length === 1 &&
     !!getProduct(sortedTransactions[0].category)?.quantity.usage;
```
szenius, line 175: Will the latest transaction always be the same policy as what is in `globalQuota`?

src/components/CustomerQuota/NoQuota/NoQuotaCard.tsx
```diff
@@ -221,8 +221,18 @@ export const NoQuotaCard: FunctionComponent<NoQuotaCard> = ({
   const showGlobalQuota =
     !!quotaResponse?.globalQuota &&
     cart.length > 0 &&
+    /**
+     * We only display global limit messages if there is only one global limit,
+     * since we have not catered for showing global limit messages for multiple
+     * categories.
+     */
+    quotaResponse.globalQuota.length === 1 &&
     !!getProduct(cart[0].category)?.quantity.usage;
```
szenius, line 230: Same question here, will the first cart item always be the same policy as what is in globalQuota?

szenius (commented):
Questions!

src/components/CustomerQuota/CheckoutSuccess/CheckoutSuccessCard.tsx
szenius: Ah, you are assuming there is only one policy for this campaign. In this case would a better check be checking that `allProducts` have length 1? Rather than checking for `globalQuota` length

src/components/CustomerQuota/CheckoutSuccess/CheckoutSuccessCard.tsx
yongjiajie: Good point, the message should still be shown if the quantity is 0. Will remove this!

src/components/CustomerQuota/CheckoutSuccess/CheckoutSuccessCard.tsx
yongjiajie: `firstGlobalQuota` sounds good!

src/components/CustomerQuota/CheckoutSuccess/CheckoutSuccessCard.tsx
yongjiajie: Good point, will change this check for both `CheckoutSuccessCard` and `NoQuotaCard`!

src/components/CustomerQuota/NoQuota/NoQuotaCard.tsx
yongjiajie: Same as above, will change this check to just check `allProducts` length.

src/components/CustomerQuota/CheckoutSuccess/CheckoutSuccessCard.tsx
szenius: Ah just to be consistent, should we also change the second condition from 

```
!!getProduct(sortedTransactions[0].category)?.quantity.usage;
```

to 

```
allProducts[0].quantity.usage
```

? 

This way, this computation also doesn't have to wait on fetching past transactions.

szenius (approved):
bitmoji

PR #190, opened 2020-09-18
fix: add additional validation for phone number identifier input using regex

dalsontws:
will we need a default regex validation? in the event phone number does not have a regex.


ilhaoni:
Haha @dalsontws should we rerun this and see whether we can merge this in this week? ^^

ilhaoni (approved):
LGTM! We'll have to add a default phone number regex in policies right haha

ilhaoni (approved):
LGTM!! Let's ship :D:D

PR #194, opened 2020-09-21
Feat/manual mrz input (ready for review)

justin0108:


src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
```diff
@@ -107,6 +115,19 @@ const CollectCustomerDetailsScreen: FunctionComponent<NavigationFocusInjectedPro
...
+
+  const getSelectionArray = useCallback((): IdentificationFlag[] => {
+    const selectionArray = [];
+    selectionArray.push(features?.id || defaultSelectedIdType);
+    features?.alternateIds &&
+      features.alternateIds.map(alternateId =>
+        selectionArray.push(alternateId)
+      );
+    return selectionArray;
+  }, [features]);
```
sebastianquek, line 130: Wonder if we want to use `useMemo` in this case since the return value should likely be the same for each campaign. Noticed that on every render of this screen, `hasMultiInputSelection` is called, which triggers this.

src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
```diff
@@ -190,6 +249,13 @@ const CollectCustomerDetailsScreen: FunctionComponent<NavigationFocusInjectedPro
           <View style={styles.headerText}>
             <AppHeader mode={config.appMode} />
           </View>
+          {hasMultiInputSelection() && (
+            <InputSelection
+              selectionArray={getSelectionArray()}
```
sebastianquek, line 254: Similarly for here, looks like we construct the array twice for both `hasMultiInputSelection` and `getSelectionArray`, useMemo should help

src/components/CustomerDetails/ManualPassportInput.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const [selectedCountry, setSelectedCountry] = useState<DropdownItem | null>(
+    null
+  );
+  const [passportNo, setPassportNo] = useState<string | null>(null);
+
+  const onItemSelection = (item: DropdownItem): void => {
+    setSelectedCountry(item);
+  };
+
+  const onPassportNoChanged = (passportNo: string): void => {
```
sebastianquek, line 46: Minor nit, shall we use `num` rather than `no`, keep reading it as "no changed" 😅

src/components/CustomerDetails/ManualPassportInput.tsx
```diff
@@ -0,0 +1,89 @@
...
+interface ManualPassportInput {
+  setIdInput: (id: string) => void;
+  submitId: () => void;
+}
+
+export const ManualPassportInput: FunctionComponent<ManualPassportInput> = ({
+  setIdInput,
+  submitId
+}) => {
+  const [selectedCountry, setSelectedCountry] = useState<DropdownItem | null>(
```
sebastianquek, line 37: Noticed this is initialised as nullable, do we need to differentiate between `null` and `undefined` in this case? will `useState<DropdownItem>()` suffice ah?

src/components/DropdownFilterModal/DropdownFilterInput.tsx
```diff
@@ -0,0 +1,76 @@
...
+  placeholder,
+  value,
+  dropdownItems,
+  onItemSelection
+}) => {
+  const [modalVisible, setModalVisible] = useState(false);
+
+  const closeModal = (): void => {
+    setModalVisible(false);
+  };
```
sebastianquek, line 46: Minor thing, noticed we extracted this but not `openModal`, do we need `openModal`?

src/components/DropdownFilterModal/DropdownFilterModal.tsx
```diff
@@ -0,0 +1,235 @@
...
+    backgroundColor: "white",
+    borderRadius: 20,
+    shadowColor: "#000",
+    shadowOffset: {
+      width: 0,
+      height: 2
+    },
+    shadowOpacity: 0.25,
+    shadowRadius: 3.84,
+    elevation: 5
```
sebastianquek, line 33: Can we use the `shadow` styling helper here?

src/components/DropdownFilterModal/DropdownFilterModal.tsx
```diff
@@ -0,0 +1,235 @@
...
+  },
+  separator: {
+    marginVertical: size(1),
+    height: 1,
+    width: "100%",
+    backgroundColor: "#CED0CE"
+  },
+  crossIconLeft: {
+    marginLeft: size(3),
+    marginTop: size(7) + 2,
```
sebastianquek, line 48: Hmmm do we need the `+ 2`? Can we use the cross icon we typically use for other modals instead?

sebastianquek (commented):
Awesome work, just a few comments!

ilhaoni (commented):
Great job! Just a few small comments too~

src/components/DropdownFilterModal/DropdownFilterModal.tsx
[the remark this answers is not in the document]
Gabrieltay: Refactored the styling and uses expo vector icon.

Gabrieltay: Change to using [cross icon](https://icons.expo.fyi/AntDesign/close)

src/utils/validatePassport.ts
```diff
@@ -0,0 +1,15 @@
+import { ERROR_MESSAGE } from "../context/alert";
+
+const validate = (inputPassport: string): boolean => {
+  const passportRegex = "^[A-Za-z]{1,3}-[a-zA-Z0-9]{5,9}$";
```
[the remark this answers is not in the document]
Gabrieltay: erm.. its because we treat this identifier like `NRIC`. Its a known identifier type unlike `STRING` where we provide the `validationRegex`. Just wonder if its weird if we have a `PASSPORT` identifier with different regex for different campaigns

package-lock.json
```diff
@@ -8772,7 +8772,8 @@
     "i18n-js": {
       "version": "3.7.1",
       "resolved": "https://registry.npmjs.org/i18n-js/-/i18n-js-3.7.1.tgz",
-      "integrity": "sha512-xrRzCeda5ZC0u0yRN+dMdidtda0N+f7t7Pek0ajWb+iyKqSGdrMmuBtbNpWJWY5N4Th0cxbp/BR57zSPdrM3Rw=="
+      "integrity": "sha512-xrRzCeda5ZC0u0yRN+dMdidtda0N+f7t7Pek0ajWb+iyKqSGdrMmuBtbNpWJWY5N4Th0cxbp/BR57zSPdrM3Rw==",
+      "dev": true
```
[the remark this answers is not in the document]
justin0108: Are you referring to the `dev` ? I saw this change in the current master branch. I think I got this change when I merge with master.. not sure why it still flag as a diff here since master already have this change

src/components/DropdownFilterModal/DropdownFilterModal.tsx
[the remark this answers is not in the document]
justin0108: I realise this is not needed as the input as been styled separately. I have removed it

src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
justin0108: good catch! I have changed to `useMemo`

src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
justin0108: have change to `useMemo`

src/components/CustomerDetails/ManualPassportInput.tsx
justin0108: done. Change to `num`

src/components/CustomerDetails/ManualPassportInput.tsx
justin0108: Yup I think `useState<DropdownItem>()` should suffice as I am not really distinguishing between `null` or `undefine` also

src/components/DropdownFilterModal/DropdownFilterInput.tsx
justin0108: Yup I have added `openModal` so that the api looks more consistent

src/components/DropdownFilterModal/DropdownFilterModal.tsx
justin0108: Have removed as per comment above

package-lock.json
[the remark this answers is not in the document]
ilhaoni: Yup yup! Ohhh, okok, hmm 🤔  okok, lets leave it first then

src/components/DropdownFilterModal/DropdownFilterInput.tsx
```diff
@@ -0,0 +1,80 @@
...
+  },
+  inputView: {
+    marginTop: size(1),
+    minHeight: size(6),
+    borderRadius: borderRadius(2),
+    borderWidth: 1,
+    backgroundColor: color("grey", 0),
+    borderColor: color("blue", 50)
+  },
+  inputText: {
```
sebastianquek, line 19: Need to add `fontFamily: brand-regular` too, right now it's the default font

justin0108: ok done.

src/utils/validatePassport.ts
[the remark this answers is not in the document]
ilhaoni: I see I see! Hmm, true, let's leave it then!

src/components/CustomerDetails/ManualPassportInput.tsx
```diff
@@ -0,0 +1,83 @@
...
+      ? setIdInput(`${selectedCountry?.id}-${passportNum}`)
+      : setIdInput("");
+  }, [selectedCountry, passportNum, setIdInput]);
+
+  return (
+    <View style={styles.centeredView}>
+      <View style={styles.inputAndButtonWrapper}>
+        <View style={styles.inputWrapper}>
+          <DropdownFilterInput
+            label="Nationality (non-SGP)"
```
ilhaoni, line 55: hmm small nit, not sure if `(non-SGP)` would be clear to the users, wonder if `(tourists)` or `(non-Singaporeans)` or `(visitors)` would be better. Can check with Eida! But doesnt have to be this PR (:

ilhaoni (approved):
LGTM! :D Thanks for the great work~~

src/components/CustomerDetails/ManualPassportInput.tsx
justin0108: Ok let me check with her again. The `non-SGP` is requested by Eida. I feedback the comments to her first

justin0108: they suggested to change to `Nationality (non-Singaporean)`

PR #195, opened 2020-09-21
fix: input env var in deploy preview

dhemutton:
This PR contains a fix for the failing deploy branch preview for storybook possibly due to an update in expo-cli requiring the env variables to be restated in the workflow job.


sebastianquek (approved):
Thanks for making the fix!

PR #198, opened 2020-09-23
feat: run e2e after staging is deployed

ddwsd:
Since the E2E tests run on push to master, it may run concurrently with the staging deployment. This could result in our E2E tests not using the newly deployed staging Expo application, which is unintended


ilhaoni (approved):
Great catch! :D 

🚀 let's go~~

PR #202, opened 2020-09-25
feat: spacing in contact number

shermainelim:
notion entry: https://www.notion.so/Spacing-in-Contact-number-62a744454055462ca8a94ba2857bd9bc

e.g. 90998822 changed to 9099 8822 with a spacing.

`INTERNATIONAL` and `NATIONAL` formats are consistent with the definition in ITU-T Recommendation E.123. However we follow local conventions such as using '-' instead of whitespace as separators. For example, the number of the Google Switzerland office will be written as "+41 44 668 1800" in `INTERNATIONAL` format, and as "044 668 1800" in `NATIONAL` format.
`E164` format is as per `INTERNATIONAL` format but with no formatting applied, e.g. "+41446681800". `RFC3966` is as per `INTERNATIONAL` format, but with all spaces and other separating symbols replaced with a hyphen, and with any phone number extension appended with ";ext=". It also will have a prefix of "tel:" added, e.g. "tel:+41-44-668-1800".

Reference: javadoc.io/doc/com.googlecode.libphonenumber/libphonenumber/8.8.10/com/google/i18n/phonenumbers/PhoneNumberUtil.PhoneNumberFormat.html

For google libphonenumber PhoneNumberUtil.format() format contains

`PhoneNumberFormat.E164`
[+65] [+6590000190]

`PhoneNumberFormat.INTERNATIONAL`
if added to screen ,becomes [+65] [+65 9000 0190]

`PhoneNumberFormat.NATIONAL`
if added to screen, [+65] [9000 0190], similar to phoneUtil.formatOutOfCountryCallingNumber(mobileNumberValue, "SG")
So this option was chosen.

`PhoneNumberFormat.RFC3966`
if added to screen, becomes tel:+65-9000 0190


shermainelim:
For Feature: Spacing in Contact Number

https://www.notion.so/701364d4fbbf4ec1801b766c589973b9?v=24aa9675206d4da9a171ca4073d81b5b&p=62a744454055462ca8a94ba2857bd9bc

src/components/Layout/PhoneNumberInput.tsx
```diff
@@ -77,7 +87,7 @@ export const PhoneNumberInput: FunctionComponent<{
         <TextInput
           style={styles.numberInput}
           keyboardType="phone-pad"
-          value={mobileNumberValue}
+          value={value}
```
sebastianquek, line 90: Can we approach this differently? Right now the `value` corresponds to 2 things: what the user sees and also what is used as the underlying state. These 2 things can actually be separate.

So the underlying state remains as what's passed in to this component (`mobileNumberValue`), but when we pass it over to `TextInput` for rendering, we transform it first,  like what you did. When the user changes the text in the input box, `onChange`  will receive the text and transform it back to the original type. In this case,  a number without the spaces.

This illustrates one key thing about frontend rendering btw, what is rendered depends on the state, and these are likely of different types. Another example is when we want to render money. We can store it in cents, but present it   in a  nice human readable way with  the dollar sign and commas. When the user edits it, we will  transform what the user has edited back to cents for storing in our state.

sebastianquek: Also,  by doing it this way, consumers of `PhoneNumberInput` don't need to know about the underlying state in it. The consumers will  just know that this renders a  phone number nicely and when a  user edits it, the  value will be feedback to the consumer as a string without spaces.

src/components/Login/LoginMobileNumberCard.tsx
```diff
@@ -61,7 +61,8 @@ export const LoginMobileNumberCard: FunctionComponent<LoginMobileNumberCard> = (
   };
 
   const onChangeMobileNumber = (text: string): void => {
-    /^\d*$/.test(text) && setMobileNumberValue(text);
+    ///^\d*$/.test(text) &&
```
sebastianquek, line 64: Do be mindful of commented out code.

In this case, as the above comment, we shouldnt need to make changes to `LoginMobileNumberCard` in this PR

sebastianquek (changes requested):

src/components/Layout/PhoneNumberInput.tsx
```diff
@@ -63,6 +64,14 @@ export const PhoneNumberInput: FunctionComponent<{
   // eslint-disable-next-line @typescript-eslint/no-empty-function
   onSubmit = () => {}
 }) => {
+  const phoneUtil = PhoneNumberUtil.getInstance();
+  const value: string =
+    mobileNumberValue.length > 2
+      ? phoneUtil.formatOutOfCountryCallingNumber(
```
sebastianquek, line 70: What's the difference between using this over `format`?

src/components/Layout/PhoneNumberInput.tsx
```diff
@@ -63,6 +64,14 @@ export const PhoneNumberInput: FunctionComponent<{
...
+          phoneUtil.parseAndKeepRawInput(mobileNumberValue, "SG"),
```
sebastianquek, line 71: Similarly, what's the difference between this over `parse`?

src/components/Layout/PhoneNumberInput.tsx
```diff
@@ -63,6 +64,14 @@ export const PhoneNumberInput: FunctionComponent<{
...
   onSubmit = () => {}
 }) => {
+  const phoneUtil = PhoneNumberUtil.getInstance();
+  const value: string =
+    mobileNumberValue.length > 2
+      ? phoneUtil.format(
+          phoneUtil.parse(mobileNumberValue, "SG"),
+          PhoneNumberFormat.NATIONAL
+        )
+      : mobileNumberValue;
```
yongjiajie, line 74: Tested this out, it threw an error when I entered a really long number, please look into it!

The error thrown was: `Error: The string supplied is too long to be a phone number`.
Have also sent you a screen recording of it!

yongjiajie (changes requested):

src/components/Layout/PhoneNumberInput.tsx
shermainelim: fix by restricting input to 8 digits.

shermainelim: @shermainelim
fix:phone number input using constant value that is found on google libphonenumber library

Reference: https://github.com/ruimarinho/google-libphonenumber/blob/master/src/phonenumberutil.js

Line 114

onChange we getting the max length of phone number from MAX_LENGTH_FOR_NSN then
then is validated. if wrong, then don't update the state.

src/components/Layout/PhoneNumberInput.tsx
```diff
@@ -63,6 +66,14 @@ export const PhoneNumberInput: FunctionComponent<{
   // eslint-disable-next-line @typescript-eslint/no-empty-function
   onSubmit = () => {}
 }) => {
+  const phoneUtil = PhoneNumberUtil.getInstance();
```
yongjiajie, line 69: Just a nitpick, maybe this can be moved outside of the scope of the component since this is a singleton? We don't need to get a new instance every time this component is rendered.

src/components/Layout/PhoneNumberInput.tsx
```diff
@@ -2,6 +2,9 @@ import React, { FunctionComponent } from "react";
 import { View, TextInput, StyleSheet } from "react-native";
 import { AppText } from "./AppText";
 import { size, color, borderRadius, fontSize } from "../../common/styles";
+import { PhoneNumberFormat, PhoneNumberUtil } from "google-libphonenumber";
+
+const MAX_LENGTH_FOR_NSN = 17;
```
yongjiajie, line 7: While this value is strictly from the library **now**, what if it changes in the future? How can we implement this in a better way? 

Can we consider leveraging the library in some way to do the _heavy lifting_?

src/components/Layout/PhoneNumberInput.tsx
yongjiajie: While the value for `MAX_LENGTH_FOR_NSN` is 17 for now, it may not be 17 in the future.

Have you considered leveraging the errors that `format` and `parse` will throw when met with an input that is too short or too long?

src/components/Layout/PhoneNumberInput.tsx
```diff
@@ -77,8 +88,13 @@ export const PhoneNumberInput: FunctionComponent<{
         <TextInput
           style={styles.numberInput}
           keyboardType="phone-pad"
-          value={mobileNumberValue}
-          onChange={({ nativeEvent: { text } }) => onChangeMobileNumber(text)}
+          value={value}
+          onChange={({ nativeEvent: { text } }) => {
```
yongjiajie, line 92: Noticed `onChange` being used but there's `onChangeText`, perhaps that is a better callback to use?
Let me know what you think!

yongjiajie (changes requested):

src/components/Layout/PhoneNumberInput.tsx
shermainelim: yes, MAX_LENGTH_FOR_NSN = 17 removed and changed to phoneNumberFormatted function

Make code cleaner with reuseable function phoneNumberFormatted and try catch to leverage on errors format and parse based on google-libphonenumber

src/components/Layout/PhoneNumberInput.tsx
shermainelim: `onChange` to `onChangeText` Because `onChangeText` use text then no need to destructure object compared to `onChange`

shermainelim: the "SG" also reformatted.

src/components/Layout/PhoneNumberInput.tsx
shermainelim: moved this to outside of scope so don't need to get new instance each time component is rendered.

src/components/Layout/PhoneNumberInput.tsx
shermainelim: yes, instead of 17 , have replaced with phoneNumberFormatted function that leverages on `format` and `parse`
for different kinds of inputs.

src/components/Layout/PhoneNumberInput.tsx
yongjiajie: Thanks!

src/components/Layout/PhoneNumberInput.tsx
yongjiajie: The change to `onChangeText` looks good!

src/components/Layout/PhoneNumberInput.tsx
yongjiajie: Thanks a lot!

src/components/Layout/PhoneNumberInput.tsx
```diff
@@ -63,6 +76,7 @@ export const PhoneNumberInput: FunctionComponent<{
   // eslint-disable-next-line @typescript-eslint/no-empty-function
   onSubmit = () => {}
 }) => {
+  const value: string = phoneNumberFormatted(mobileNumberValue);
```
yongjiajie, line 79: Have you considered using `useState` to maintain this value since it is being altered in this component?

This approach works now because `onChangeMobileNumber` in the parent component actually updates the value and passes it back down, which reinstaniates this `const` with the new value.

Instantiating this as `const` also makes it such that this value is immutable, which is not the case in this component.

Using `useState` will allow us to easily manipulate this value in the component, as well as leverage React's state management if need be.

src/components/Layout/PhoneNumberInput.tsx
```diff
@@ -2,6 +2,19 @@ import React, { FunctionComponent } from "react";
...
 import { size, color, borderRadius, fontSize } from "../../common/styles";
+import { PhoneNumberFormat, PhoneNumberUtil } from "google-libphonenumber";
+
+const singaporeNumberCode = "SG";
+const phoneUtil = PhoneNumberUtil.getInstance();
+
+const phoneNumberFormatted = (mobileNumberValue: string): string => {
+  return mobileNumberValue.length >= 2
+    ? phoneUtil.format(
+        phoneUtil.parse(mobileNumberValue, singaporeNumberCode),
```
yongjiajie, line 13: This is another way of declaring default values, without having to instantiate a new `const` above, as well as allowing extensibility in the future if necessary:

```suggestion
const phoneNumberFormatted = (mobileNumberValue: string, region: string = "SG"): string => {
  return mobileNumberValue.length >= 2
    ? phoneUtil.format(
        phoneUtil.parse(mobileNumberValue, region),
```

src/components/Layout/PhoneNumberInput.tsx
```diff
@@ -71,14 +85,20 @@ export const PhoneNumberInput: FunctionComponent<{
...
-          value={mobileNumberValue}
-          onChange={({ nativeEvent: { text } }) => onChangeMobileNumber(text)}
+          value={value}
+          onChangeText={text => {
+            try {
+              const cleanText: string = text.replace(" ", "");
+              phoneNumberFormatted(text);
+              onChangeMobileNumber(cleanText);
+            } catch {}
+          }}
```
yongjiajie, line 101: Perhaps this can be extracted into a function that does the following:

- Formats the phone number
- Prevents input of more than the maximum length
- Allows input of less than 2 without hard coding `length <= 2` (what if the library decides 2 is not short tomorrow?)
- Removes formatting from the formatted phone number and passes it to `onChangeMobileNumber`

```suggestion
          onChangeText={text => onHandleChangeText(text)}
```

yongjiajie (changes requested):

src/components/Layout/PhoneNumberInput.tsx
shermainelim: but if `useState` then the regex in loginMobileNumberCard will fail. Have initially considered using `useState` to modify the formatted value in the mobileNumberValue.

but this approach will require cleaning up the data when submitting the data in the loginMobileNumberCard. For example, removing the space, then have to modify in loginMobileNumberCard.

src/components/Layout/PhoneNumberInput.tsx
shermainelim: added onHandleChangeText 
- Formats the phone number
- Input length not restricted as to fit International Numbers next time
-Removed "SG" and used getRegionCodeForCountryCode so can fit formatting for International Numbers

src/components/Layout/PhoneNumberInput.tsx
shermainelim: -Removed "SG" and used getRegionCodeForCountryCode so can fit formatting for International Numbers

src/components/Layout/PhoneNumberInput.tsx
shermainelim: @yongjiajie  Will like to check if `LoginMobileNumberCard` from `line 48-50`:
  const [countryCode, setCountryCodeValue] = useState(
    config.fullMobileNumber?.countryCode ?? "+65"
  );

If this is at +65, if I try to type other Country Code & Number in PhoneNumberInput, will the formatting prevent me?

yongjiajie: No it won't, the value within `useState` is just the initial value.

src/components/Layout/PhoneNumberInput.tsx
```diff
@@ -63,6 +66,31 @@ export const PhoneNumberInput: FunctionComponent<{
...
+  };
+  const [value, setValue] = useState<string>(
+    formatPhoneNumber(mobileNumberValue.replace(" ", ""))
+  );
+  useEffect(() => {
+    if (value) {
+      onChangeMobileNumber(value.replace(" ", ""));
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [value]);
```
yongjiajie, line 93: I didn't think of exhaustive deps, let's change this back to how you did it before:

```suggestion
  const onHandleChangePhoneNumber = (mobileNumberValue: string) => {
    const formattedPhoneNumber = formatPhoneNumber(mobileNumberValue);
    onChangeMobileNumber(formattedPhoneNumber);
    setMobileNumber(formattedPhoneNumber);
  }
```

src/components/Layout/PhoneNumberInput.tsx
```diff
@@ -63,6 +66,31 @@ export const PhoneNumberInput: FunctionComponent<{
...
+        PhoneNumberFormat.NATIONAL
+      );
+    } catch (e) {
+      if (value && value?.length < mobileNumberValue.length) {
+        return value;
+      }
+      return mobileNumberValue;
+    }
+  };
+  const [value, setValue] = useState<string>(
```
yongjiajie, line 85: Just a nitpick, can we rename this to `mobileNumber` and `setMobileNumber`?

src/components/Layout/PhoneNumberInput.tsx
```diff
@@ -71,14 +99,14 @@ export const PhoneNumberInput: FunctionComponent<{
...
+          onChangeText={text => onChangeCountryCode(text)}
         />
         <AppText style={styles.hyphen}>-</AppText>
         <TextInput
           style={styles.numberInput}
           keyboardType="phone-pad"
-          value={mobileNumberValue}
-          onChange={({ nativeEvent: { text } }) => onChangeMobileNumber(text)}
+          value={value}
+          onChangeText={text => setValue(formatPhoneNumber(text))}
```
yongjiajie, line 109: Change this back to `onHandleChangePhoneNumber(text)`, I don't think we should rely on `useEffect` for this, sorry!

yongjiajie (changes requested):
I have some changes to request for after thinking about it, sorry!

src/components/Layout/PhoneNumberInput.tsx
```diff
@@ -63,6 +66,33 @@ export const PhoneNumberInput: FunctionComponent<{
...
+      }
+      return mobileNumberValue;
+    }
+  };
+  const [mobileNumber, setMobileNumber] = useState<string>(
+    formatPhoneNumber(mobileNumberValue.replace(" ", ""))
+  );
+
+  // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+  const onHandleChangePhoneNumber = (text: string) => {
```
yongjiajie, line 90: I would try to avoid disabling `eslint` rules and only really disable them if you cannot easily fix it:
```suggestion
  const onHandleChangePhoneNumber = (text: string): void => {
```

yongjiajie (changes requested):
Just one final change!

src/components/Layout/PhoneNumberInput.tsx
yongjiajie: Thanks a lot Shermaine!

src/components/Layout/PhoneNumberInput.tsx
yongjiajie: Thanks for changing this Shermaine!

src/components/Layout/PhoneNumberInput.tsx
yongjiajie: 👍

yongjiajie (approved):
Looks good to me!

src/components/Layout/PhoneNumberInput.tsx
```diff
@@ -63,6 +66,32 @@ export const PhoneNumberInput: FunctionComponent<{
...
+    } catch (e) {
+      if (mobileNumber && mobileNumber?.length < mobileNumberValue.length) {
+        return mobileNumber;
+      }
+      return mobileNumberValue;
+    }
+  };
+  const [mobileNumber, setMobileNumber] = useState<string>(
+    formatPhoneNumber(mobileNumberValue.replace(" ", ""))
+  );
```
yongjiajie, line 87: Remove this:
```suggestion
```

src/components/Layout/PhoneNumberInput.tsx
```diff
@@ -71,14 +100,14 @@ export const PhoneNumberInput: FunctionComponent<{
...
-          onChange={({ nativeEvent: { text } }) => onChangeCountryCode(text)}
+          onChangeText={text => onChangeCountryCode(text)}
         />
         <AppText style={styles.hyphen}>-</AppText>
         <TextInput
           style={styles.numberInput}
           keyboardType="phone-pad"
-          value={mobileNumberValue}
-          onChange={({ nativeEvent: { text } }) => onChangeMobileNumber(text)}
+          value={mobileNumber}
```
yongjiajie, line 109: Some more cleaning up:

```suggestion
          value={formatPhoneNumber(mobileNumberValue)}
```

src/components/Layout/PhoneNumberInput.tsx
```diff
@@ -63,6 +66,32 @@ export const PhoneNumberInput: FunctionComponent<{
...
+  const [mobileNumber, setMobileNumber] = useState<string>(
+    formatPhoneNumber(mobileNumberValue.replace(" ", ""))
+  );
+
+  const onHandleChangePhoneNumber = (text: string): void => {
+    const newMobileValue = formatPhoneNumber(text);
+    setMobileNumber(newMobileValue);
+    const cleanText: string = newMobileValue.replace(" ", "");
+    onChangeMobileNumber(cleanText);
+  };
```
yongjiajie, line 94: Just some cleaning up:

```suggestion
  const onHandleChangePhoneNumber = (text: string): void => {
    const cleanText: string = formatPhoneNumber(text).replace(" ", "");
    onChangeMobileNumber(cleanText);
  };
```

yongjiajie (changes requested):
After discussion with @sebastianquek we should let the parent handle the state while this component handles the display! So we will have to remove some parts!

sebastianquek:
@shermainelim I noticed you reverted many changes :O could you please rebase this PR again?

PR #203, opened 2020-09-25
fix: mask sensitive data on sentry

ddwsd:
Mask sensitive data on sentry!

### DIFF
--- a/src/hooks/usePastTransaction/usePastTransaction.tsx
+++ b/src/hooks/usePastTransaction/usePastTransaction.tsx
@@ -35,7 +35,7 @@ export const usePastTransaction = (
       } catch (error) {
         Sentry.captureException(
           `Unable to fetch past transactions: ${ids.map(id =>
-            id.replace(/^[\w]{5}/g, "*****")
+            id.slice(-4).padStart(id.length, "*")
           )}`
         );
         setError(new Error(ERROR_MESSAGE.PAST_TRANSACTIONS_ERROR));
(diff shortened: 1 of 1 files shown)


src/hooks/usePastTransaction/usePastTransaction.tsx
```diff
@@ -33,7 +33,11 @@ export const usePastTransaction = (
         );
         setPastTransactionsResult(pastTransactionsResponse?.pastTransactions);
       } catch (error) {
-        Sentry.captureException(`Unable to fetch past transactions: ${ids}`);
+        Sentry.captureException(
+          `Unable to fetch past transactions: ${ids.map(id =>
+            id.replace(/\w/g, "*")
```
szenius, line 38: Do we want to leave the last 4 chars for debugging?

ddwsd: for nrics it doesn't make a difference i think, but yeah for the other types of identification

szenius: What do you mean it doesn't make a difference ah

szenius: Btw in the backend we use `string.padStart()` to do masking, maybe we want to be consistent with that~ Can check out https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart

ddwsd: nice, thanks! was wondering what's the best way lol. 
i think the nrics would be masked in our backend and db anyway, so it's a bit harder to debug

szenius (approved):
bitmoji

PR #204, opened 2020-09-27
feat: add operatorToken as data when sentry capture exceptions

dalsontws:


src/components/CampaignInitialisation/CampaignInitialisationScreen.tsx
```diff
@@ -105,6 +105,11 @@ export const CampaignInitialisationScreen: FunctionComponent<NavigationProps> =
     if (updateCampaignConfigError) {
       if (updateCampaignConfigError instanceof CampaignConfigError) {
         Sentry.captureException(updateCampaignConfigError);
+        Sentry.addBreadcrumb({
+          data: {
+            operatorToken: authCredentials.operatorToken
+          }
+        });
```
ilhaoni, line 112: haha I suspect we should add the breadcrumb before `captureException` instead haha 😆

src/components/CustomerAppeal/CustomerAppealScreen.tsx
```diff
@@ -112,6 +112,11 @@ export const CustomerAppealScreen: FunctionComponent<NavigationProps> = ({
     );
     if (appealProduct === undefined) {
       Sentry.captureException(`Unable to find appeal product: ${productName}}`);
+      Sentry.addBreadcrumb({
+        data: {
+          operatorToken
+        }
+      });
```
ilhaoni, line 119: Same for this 😆

src/components/Login/LoginContainer.tsx
```diff
@@ -190,6 +190,11 @@ export const InitialisationContainer: FunctionComponent<NavigationProps> = ({
         if (!RegExp(DOMAIN_FORMAT).test(queryEndpoint)) {
           const error = new Error(`Invalid endpoint: ${queryEndpoint}`);
           Sentry.captureException(error);
+          Sentry.addBreadcrumb({
+            data: {
+              operatorToken: queryKey
+            }
+          });
```
ilhaoni, line 197: and this

ilhaoni (changes requested):
Hmm, I think the breadcrumbs added in this PR will have to be added before `Sentry.captureException` haha, so that it can be added before Sentry throws the error~ 😆 

Another thought that I have, we currently have `useEffect()` functions that add breadcrumb like
```
  useEffect(() => {
    Sentry.addBreadcrumb({
      category: "navigation",
      message: "CampaignLocationsScreen"
    });
  }, []);
```
and 
```
  useEffect(() => {
    Sentry.addBreadcrumb({
      category: "loginStage",
      message: loginStage
    });
  }, [loginStage]);
```

Just wondering if it makes more sense to add `data: { operatorToken }` there instead haha.

What are your thoughts haha~ \^^/

PR #207, opened 2020-09-30
feat: Code Scanner Guidance and Precision

yongjiajie:
[Notion Reference](https://www.notion.so/Code-Scanner-Guidance-and-Precision-96edd24c72cd4c52a1cb893a8b78ce1a)

This feature introduces the following features:
- Removed cancel button at the bottom in favor of top-left back button
- Added light box to scanner, scanning only occurs when code is in focused area
- Added label to light box informing user what type of code to scan
- Chinese translations for the code scanner (see [this](https://www.notion.so/Chinese-translation-for-app-Scanner-Guidance-and-Precision-78003762c891412e83cb38d0426fb328))

The scanner should appear in the following pages:
- `CollectCustomerDetailsScreen`
- `AddUserModal`
- `IdentifierScanModal`
-  `LoginContainer`


src/components/IdScanner/IdScanner.tsx
```diff
@@ -1,63 +1,121 @@
...
+};
+
 export type Camera = Pick<
   BarCodeScannerProps,
   "onBarCodeScanned" | "barCodeTypes"
->;
+> & {
+  cancelButtonText?: string;
+  onCancel?: () => void;
+  hasLimitedInterestArea?: boolean;
```
millasml, line 64: I think hasLimitedInterestArea is not necessary, you can just check if interestAreaLayout is there!

millasml (changes requested):

src/components/IdScanner/IdScanner.tsx
yongjiajie: I have removed `hasLimitedInterestArea` here!

dalsontws (changes requested):
not sure if its just me, noticed this when trying to use the scanner!
IMG_7018

src/common/i18n/translations/en.ts
```diff
@@ -33,7 +33,10 @@ export const en: Translations = {
     enterManually: "Enter manually",
     enterVoucherId: "Enter voucher ID",
     scanToCheck: "Scan to check validity",
-    continueScanning: "Continue scanning"
+    continueScanning: "Continue scanning",
+    scanBarCode: "Scan Barcode",
```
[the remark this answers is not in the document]
yongjiajie: I agree with you, have renamed this to `scanBarcode`.

The text should be "Scan Barcode" as per our documentation [here](https://www.notion.so/Code-Scanner-Guidance-and-Precision-96edd24c72cd4c52a1cb893a8b78ce1a)!

yongjiajie:
> not sure if its just me, noticed this when trying to use the scanner!
> IMG_7018

Looks like the way I was storing and retrieving the labels was such that we couldn't retrieve the translations before it was stored, have fixed this!

src/components/IdScanner/IdScanner.tsx
```diff
@@ -71,21 +150,51 @@ export const IdScanner: FunctionComponent<IdScanner> = ({
...
+        >
+          <View style={styles.backButtonWrapper}>
+            <TransparentButton
+              onPress={onCancel}
+              text={cancelButtonText}
+              icon={
+                <Ionicons
+                  name="ios-arrow-back"
+                  size={size(2)}
+                  color={color("grey", 0)}
```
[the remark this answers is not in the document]
yongjiajie: I'm following the documentation [here](https://www.notion.so/Code-Scanner-Guidance-and-Precision-96edd24c72cd4c52a1cb893a8b78ce1a)! Will double check with her too, thanks for the heads up!

dalsontws (changes requested):
hmm i seem to be having an issue scanning the IC and QR, having this error: 
IMG_7036

tested this on pod staging, and also noticed sometimes clicking on `Scan identification` results in a "scan QR" instead of "scan barcode", sent overt to you on slack!

src/components/IdScanner/IdScanner.tsx
```diff
@@ -124,10 +125,12 @@ export const IdScanner: FunctionComponent<IdScanner> = ({
...
 }) => {
+  const [platform] = useState<string>(Platform.OS);
   const [hasCameraPermission, setHasCameraPermission] = useState(false);
-  const interestAreaLayout = hasLimitedInterestArea
-    ? getInterestAreaDimensions(barCodeTypes)
-    : undefined;
+  const interestAreaLayout =
+    hasLimitedInterestArea && platform === "android"
+      ? getInterestAreaDimensions(barCodeTypes)
+      : undefined;
```
dalsontws, line 133: hmm not sure if this is the issue here, but currently im unable to successfully scan IDs or QRs. like the scanner no longer working

dalsontws: also to clarify! so ios will not have any labels/guidance right? i.e. same as previously!

dalsontws (changes requested):

src/components/IdScanner/IdScanner.tsx
yongjiajie: The issue was actually another check below, fixed it!

For now, the decision is for iOS devices to not have any guidance. However, moving forward we should have labels for both iOS and Android. We're pending advice from our UX designers for the labels!

dalsontws (approved):
Tested on iOS for scanning of ID and QR, works, LGTM!

src/components/IdScanner/IdScanner.tsx
dalsontws: okay thanks for clarifying! for the labels, will be included in this PR?

yongjiajie: Yes, the labels will be included in this PR!

yongjiajie:
After discussion, the final decision for guidance and precision will be as follows:
- Both Android and iOS will have a lightbox with opacity (i.e. users will be able to see the background with opacity)
- **ONLY** Android will have scanning limited to the lightbox boundaries.

dalsontws (approved):
Tested on iOS. 
- Translucent background
- Labels for Barcode and QR
- Can scan codes even when outside lightbox

LGTM!

millasml (approved):

PR #214, opened 2020-10-06
feat: translation for campaign-specific copies

szenius:
This PR adds translations for campaign-specific copies. 
* Frontend passes backend `language` parameter in request body based on i18n.locale
* Backend returns c13n object with translations
* When displaying strings, check c13n for the corresponding translation

---

- [x] Updated fetch from backend to include `language` in request body and `c13n` hash in request and reponse bodies
- [x] Added `useTranslate` hook to abstract away translation work
- [x] Added translation for dynamic copies in components
- [x] Added and updated unit tests

### Testing instructions
* Test against [PR158](https://github.com/rationally-app/backend-api/pull/158) before it is merged
* Test against [PR163](https://github.com/rationally-app/backend-api/pull/163) for merchant flow


szenius:
There is an error that pops up on checkout for merchant flow, still looking into that, but since the PR is large and takes some time to go through + most parts are not related, I think can start reviewing le

src/hooks/useTranslate/useTranslate.ts
```diff
@@ -0,0 +1,66 @@
...
+    return (c13nStore && key && c13nStore[key]) ?? key;
+  };
+
+  const c13ntForUnit = (
+    unit: CampaignPolicy["quantity"]["unit"]
+  ): CampaignPolicy["quantity"]["unit"] =>
+    unit && {
+      ...unit,
+      label: c13nt(unit.label)
+    };
```
yongjiajie, line 61: Why is it done this way specifically for `unit`? Is it because it is nested within `quantity.unit`?

yongjiajie (commented):
Just a single question, everything looks good after adding translations and testing it out!

src/hooks/useTranslate/useTranslate.ts
szenius: Yeah, and `unit` can also be undefined. It's either this or every where that uses unit, i do 

```
const unit = unit && {
      ...unit,
      label: c13nt(unit.label)
    }
```

which seems like a lot of duplication.

yongjiajie: I see, was just curious. I was thinking of whether to generalize this, but let's leave that for another time!

szenius: Np, good to question :)

yongjiajie (approved):
Looks good!

PR #215, opened 2020-10-06
feat: redemption stats ui changes

dalsontws:
[notion](https://www.notion.so/Disabled-chevrons-for-future-date-in-redemption-statistics-and-sorting-will-be-by-sort-property-b3226495c3b34d4fa7eea588c5d6fc12)

Completing certain UI changes that were not mentioned/confirmed in previous story!

- [x]  Ensure stats order is standardised with policy order property
- [x]  Disable right chevron if date is > today
- [x]  add `via appeal` instead of `*chargeable` , applied to all appeal tt token items

### DIFF
--- a/src/hooks/useDailyStatistics/utils.ts
+++ b/src/hooks/useDailyStatistics/utils.ts
@@ -8,6 +8,7 @@ type SummarisedTransactions = {
     name: string;
     category: string;
     quantityText: string;
+    descriptionAlert?: string;
   }[];
   summarisedTotalCount: number;
 };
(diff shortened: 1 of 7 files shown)


src/hooks/useDailyStatistics/utils.ts
```diff
@@ -40,7 +42,13 @@ export const countTotalTransactionsAndByCategory = (
...
               label: " qty"
             }
-          )
+          ),
+          descriptionAlert:
+            policies?.find(item => item.category === key)?.categoryType ===
+            "APPEAL"
+              ? "via appeal"
+              : undefined,
+          order: policies?.find(item => item.category === key)?.order ?? -1
```
yongjiajie, line 51: Perhaps you can define `policies?.find(item => item.category === key)` in a `const` above and then refer to it's properties here, looks like you're repeating it a lot!

yongjiajie (changes requested):

src/hooks/useDailyStatistics/utils.ts
dalsontws: good call, changed this to `findItemByCategory`!

yongjiajie (approved):
Great work! 🎉

PR #216, opened 2020-10-07
chore: added i18n helper

yongjiajie:
Added helpers to leverage TypeScript when displaying i18n strings:

- `i18nString`
- `i18nErrorString`

Refactored all existing `i18n.t()` to use this new helper.

**To Do**

- [x] Fixed `NotEligibleCard` title not being translated
- [x] Updated `CampaignInitialisationScreen` modals not being translated
- [ ] Update Translation technical document to reflect using these new helpers


src/utils/i18nString.ts
```diff
@@ -0,0 +1,27 @@
...
+  U extends keyof Translations[T]
+>(
+  componentName: T,
+  componentString: U,
+  dynamicContent?: Record<string, string | number>
+): string => {
+  return i18n.t(`${componentName}.${componentString}`, dynamicContent);
+};
+
+export const i18nErrorString = <
```
szenius, line 15: Why must have separate utility function for this?

src/utils/i18nString.ts
```diff
@@ -0,0 +1,27 @@
+import i18n from "i18n-js";
+import { Translations } from "../common/i18n/translations/type";
+
+export const i18nString = <
```
szenius, line 4: Using a verb might help to describe better what this does. E.g. `getTranslatedStringWithI18n`

src/components/CampaignInitialisation/UpdateFromAppStoreAlert.tsx
```diff
@@ -23,10 +24,17 @@ export const UpdateFromAppStoreAlert: FunctionComponent = () => {
   return (
     <AlertModal
       alertType="INFO"
-      title="Outdated app"
-      description={`Update your app through the ${storeName}.`}
+      title={i18nErrorString("outdatedAppUpdate", "title")}
```
szenius, line 27: Good catch!

src/utils/i18nString.ts
yongjiajie: Have refactored this to be a single utility function!

src/utils/i18nString.ts
yongjiajie: I agree with this, have refactored it.

src/components/CampaignInitialisation/UpdateFromAppStoreAlert.tsx
yongjiajie: Thank you! 👍

src/utils/i18nString.ts
sebastianquek: just noticed this, since we're going to be using this everywhere, can we keep the fn name short?

yongjiajie: After some deliberation and discussion, we have decided this will be named `i18nt`!

src/components/CampaignLocations/CampaignLocationsListItem.tsx
```diff
@@ -63,10 +63,10 @@ export const CampaignLocationsListItem: FunctionComponent<
           ]}
         >
           {hasExpired
-            ? i18n.t("customerQuotaScreen.campaignExpiredOn", {
+            ? i18nt("customerQuotaScreen", "campaignExpiredOn", undefined, {
```
szenius, line 66: Oooh, actually rather than doing this `undefined` thing, what if `i18n` took in two arguments

```
i18n(["customerQuotaScreen", "campaignExpiredOn", ...], {variables})
```

First array can be infinite size, so we can be flexible for all levels of nesting. What do you think?

szenius: In that case this can be an array of two strings rather than having `undefined`

szenius (commented):
suggestion~~

src/utils/translations.ts
```diff
@@ -0,0 +1,42 @@
...
+  dynamicContent?: Record<string, string | number>
+): string {
+  if (!errorMessageContent) {
+    return i18n.t(`${componentName}.${componentString}`, dynamicContent);
+  } else {
+    return i18n.t(
+      `${componentName}.${componentString}.${errorMessageContent}`,
+      dynamicContent
+    );
+  }
```
szenius, line 41: What about 

```javascript
return i18n.t(
      `${componentName}.${componentString}${errorMessageContent ? 
`.${errorMessageContent}` : ""}`,
      dynamicContent
    );
```

szenius (changes requested):
Various screens having problems
* undefined Limit reached title
* Cannot purchase/redeem has weird new lines
* Campaign list item missing validity/expiry date
* No quota screen has english full stop for chinese text

Sent screens on Slack, adding this review for future reference

src/components/CampaignLocations/CampaignLocationsListItem.tsx
yongjiajie: After much discussion and deliberation, we have decided this may not be worth the effort to fix now, since it will take a much more complicated solutions than necessary to both retain the type checking and have a succinct function signature.

src/utils/translations.ts
yongjiajie: This is done!

src/components/CustomerQuota/ItemsSelection/ItemNoQuota.tsx
```diff
@@ -48,10 +48,8 @@ export const ItemNoQuota: FunctionComponent<{
       <View style={styles.feedbackWrapper}>
         <AppText style={styles.feedbackText}>
           {type === "REDEEM"
-            ? i18n.t("notEligibleScreen.notEligible")
-            : `${i18n.t("notEligibleScreen.cannot")}\n${i18n.t(
-                "notEligibleScreen.purchase"
-              )}`}
+            ? i18nt("notEligibleScreen", "notEligible")
```
szenius, line 51: Wasn't this "Cannot redeem" ah

szenius (approved):
image

PR #217, opened 2020-10-07
hotfix: empty ErrorBoundaryContent 

yongjiajie:
The `ErrorBoundaryContent` was found to be empty, since the wrong translation keys were referenced.
This hotfix fixes that.


sebastianquek (approved):
👍

PR #218, opened 2020-10-09
Chore: validation before payment

millasml:
for this ticket: https://www.notion.so/701364d4fbbf4ec1801b766c589973b9?v=24aa9675206d4da9a171ca4073d81b5b&p=1026500a7a604ca2a209e69fd64325a1

Using backend-api PR 151


src/components/CustomerQuota/ItemsSelection/ItemsSelectionCard.tsx
```diff
@@ -152,6 +161,13 @@ export const ItemsSelectionCard: FunctionComponent<ItemsSelectionCard> = ({
         setIsVisible={setIsAddUserModalVisible}
         validateAndUpdateIds={onCheckAddedUsers}
       />
+      <PaymentConfirmationModal
+        isVisible={showPaymentAlert}
+        commitCart={commitCart}
+        cancelPayment={() => {
+          cancelCart(() => setShowPaymentAlert(false));
+        }}
+      />
```
yongjiajie, line 170: You could look into replacing this with `showConfirmationAlert`, you could alter it to toggle visibility if you need that functionality.

yongjiajie (changes requested):
Just some minor changes from a cursory look, will look through more thoroughly after your run through with us!

src/components/CustomerQuota/ItemsSelection/ItemsSelectionCard.tsx
millasml: I tried doing this but it was not possible! My attempt is [here in this commit](https://github.com/rationally-app/mobile-application/pull/218/commits/99704b7302b7e752dcf5498eb055a6d56197f121#diff-d40004bc8e92ebfa7cb6f781d83fecc05dc6b4a26e3d06a73107f0971184f750R141)

The problem is when I pass `commitCart` as an argument in the `showConfirmationAlert` function, all the states get captured. Because `showConfirmationAlert()` is called before `reserveCart` is done executing (due to the async nature of js), the state that is captured in `commitCart` is "RESERVING", and we do not get the correct `checkoutResult`.

I tried awaiting `reserveCart` but that didn't work either, because it is wrapped in `useCallback`

yongjiajie (changes requested):
The tests look good, just have to remove some extra stuff!

szenius:
@yongjiajie Is this PR still relevant?

yongjiajie:
No, we have a new PR that implements this feature already done by @andrehadianto! We can close this.

PR #219, opened 2020-10-10
fix: Trigger Slack Report when tests fail

younglim:
Currently, without `if: always()` specified at certain steps in the workflow, reports and notifications will only be triggered if tests pass (exit code = 0).  Report when tests fail (exit code > 0).


yongjiajie (approved):
Looks good!

PR #223, opened 2020-10-13
Revert "chore: useQuota hook"

yongjiajie:
Reverts rationally-app/mobile-application#165.

A bug was found where the appeal flow was affected.
`hasNoQuota()` in the `useQuota` hook will have to be refactored to take into account the new quota response, as well as the appeal flow.


ilhaoni (approved):
Looks good! Thanks! :D

PR #224, opened 2020-10-13
chore: useQuota Hook

yongjiajie:
https://www.notion.so/701364d4fbbf4ec1801b766c589973b9?v=24aa9675206d4da9a171ca4073d81b5b&p=2c332462475940f0884703d9ba4fb8a4

Creating a separate useQuota hook out of the useCart hook.

- [x] create useQuota hook
- [x] update useCart to remove all quota related things (as much as possible)
- [x] fix tests for useCart
- [x] move some test to useQuota
- [x] update codebase to use useQuota
- [x] To ensure that appeal flow works

This was previously merged, but a bug was found such that the appeal flow did not work as intended.
`hasNoQuota` in `useQuota` hook has to be refactored to take into consideration appeal flow products.

**Potentially Breaking Changes**
- `@testing-library/react-hooks` is updated from `3.2.1` to `3.4.2`, which changes how `rerender` works.


yongjiajie:
The only change from the previous `useQuota` hook was that we checked for the quotas using filtered quotas instead of all quotas. All quotas contained appeal products, which had a non-zero quantity, which led to the appeal flow failing.

src/hooks/useQuota/useQuota.test.tsx
```diff
@@ -442,6 +536,50 @@ describe("useQuota", () => {
...
+            products: defaultAppealProducts
+          }
+        }
+      );
+      await waitForNextUpdate();
+
+      expect(result.current.quotaState).toStrictEqual("DEFAULT");
+    });
+
+    it("should set quota state to NO_QUOTA if non-appeal products quantities are zero", async () => {
```
sebastianquek, line 561: Should this be `if appeal products quantities are zero`?

yongjiajie: In this test, I am checking that the quota state is determined only by non-appeal products.
Let me know if the test name or variable names in the tests are confusing!

sebastianquek (approved):
LGTM!

PR #225, opened 2020-10-14
fix clearing id when switching tab

justin0108:


ilhaoni (approved):
Tested and it's fixed! :D Thanks for fixing this~ \^^/

PR #226, opened 2020-10-19
feat: appeal flow updates

shermainelim:
Notion Reference: [Link](https://www.notion.so/Yuhua-Appreciation-Gift-deployment-352aa506aebf463e9e07f9ee999097c1)

This PR is related to updates to appeal flow for both front-end and Yuhua Campaign. 

- [x] `Raise appeal` button showing up even when appeal products do not have quantity
         - Fixed: Check whether the quota category is for an appeal product, then check its quantity
- [x] Replace `NotEligibleCard` with new text (both English and Chinese)
- [x] Remove spacing in limit reached title:


yongjiajie (changes requested):
As spoken, please avoid hard-coding values as it will introduce tech debt. 

Let's move forward with a solution where `AppealButton` is only shown if:

1. The policy contains appeal products
2. There is a non-zero quota remaining for appeal products

yongjiajie (changes requested):
Let's pass the `allQuotaResponse` from `CustomerQuotaScreen` rather than calling the hook here, since the hook is already used in the parent screen `CustomerQuotaScreen`.

With regards to `hasAppealProduct`, your implementation will result in the `AppealButton` showing up even when there are no appeal products in the policy. We want to check that (1) there are appeal products, and (2) that these appeal products have a non-zero quantity.

(1) can be achieved by filtering through `allProducts` on `categoryType === "APPEAL"`.

(2) can be achieved by storing the list of `category` in (1) and then filtering `allQuotaResponse` and checking the `quantity`.

yongjiajie (changes requested):

yongjiajie (changes requested):
Looks good!

ilhaoni:
haha, do try to use the PR title to show the purpose of the PR. rather than saying its Yuhua Appreciation Gift haha. I think something better would be

"feat: show appeal button when there's appeal products and available quota"

what do yall think?

src/common/i18n/translations/en.ts
```diff
@@ -86,10 +86,10 @@ export const en: Translations = {
     previouslyRedeemed: "Previously redeemed",
     showMore: "Show more",
     showLess: "Show less",
-    today: "for today",
-    limitReachedRecent: "Limit reached %{time} ago %{today}.",
-    limitReachedDate: "Limit reached on %{dateTime} %{today}.",
-    limitReached: "Limit reached %{today}.",
+    today: " for today",
+    limitReachedRecent: "Limit reached %{time} ago%{today}.",
```
ilhaoni, line 90: I think there's supposed to be a space between `ago` and `%{today}`?

src/common/i18n/translations/en.ts
```diff
@@ -106,7 +106,8 @@ export const en: Translations = {
   },
   notEligibleScreen: {
     notEligible: "Not eligible",
-    logAppeal: "Please log an appeal request.",
+    logAppeal:
+      "Not eligible to collect any item. Contact your in-charge to find out about the appeal guideline.",
```
ilhaoni, line 110: Small nit, guideline or guidelines?

ilhaoni (changes requested):
Thanks for the work! (: 

Some changes to be made. This PR contains updates to appeal flow, not specific to Yuhua campaign. 

Please change the Pr name to be "feat: appeal flow updates"

The description of the PR doesn't look correct also, maybe it wasn't updated after the new commits. Let's change it to reflect the 3 main updates to the appeal flow instead. So that overall it explains what were the changes done. (:

src/common/i18n/translations/en.ts
shermainelim: changed to guidelines

src/common/i18n/translations/en.ts
shermainelim: added space

shermainelim:
> Thanks for the work! (:
> 
> Some changes to be made. This PR contains updates to appeal flow, not specific to Yuhua campaign.
> 
> Please change the Pr name to be "feat: appeal flow updates"
> 
> The description of the PR doesn't look correct also, maybe it wasn't updated after the new commits. Let's change it to reflect the 3 main updates to the appeal flow instead. So that overall it explains what were the changes done. (:

Updated PR name and PR description too.

src/common/i18n/translations/en.ts
yongjiajie: @ilhaoni We moved the space over the `today` because `%{today}` is only shown in certain campaigns, and adding a space there would introduce a space in campaigns where `%{today}` is not shown.

@shermainelim Please make sure that all the `limitReached` translation copies do not have a space before `%{today}` and that `today` has a space before the copy like so: `" for today"`.

ilhaoni: ah ok, thanks @yongjiajie for explaining!

@shermainelim sorry I want to check in, did this slipped your mind, or you're not really sure about the reason of why we removed the space? This is currently just a small display change, so it's no biggie, but I do wanna highlight that in software engineering, reviewers will ask questions about the PR, and we do expect the writers of the PR to be able to explain the changes that was made.

ilhaoni: if it's a valid change, then no changes will be required. then both reviewers and writer will learn also.

src/components/CustomerQuota/NoQuota/NoQuotaCard.tsx
```diff
@@ -214,8 +216,18 @@ export const NoQuotaCard: FunctionComponent<NoQuotaCard> = ({
     ? differenceInSeconds(now, latestTransactionTime)
     : -1;
 
+  const appealProductsCategories = allProducts
+    ?.filter((product) => product.categoryType === "APPEAL")
+    .map((product) => product.category);
+
   const hasAppealProduct =
-    allProducts?.some((policy) => policy.categoryType === "APPEAL") ?? false;
+    (appealProductsCategories &&
```
millasml, line 224: I think you should add a few test cases to test the functionality of this. Maybe move this out of NoQuotaCard, and then test it in `NoQuotaCard.test.ts`?

see the function `getLatestTransactionTime` to see what I mean!

applies to appealProductCategories too

millasml (changes requested):

src/common/i18n/translations/en.ts
```diff
@@ -86,7 +86,7 @@ export const en: Translations = {
     previouslyRedeemed: "Previously redeemed",
     showMore: "Show more",
     showLess: "Show less",
-    today: "for today",
+    today: " for today",
     limitReachedRecent: "Limit reached %{time} ago %{today}.",
     limitReachedDate: "Limit reached on %{dateTime} %{today}.",
     limitReached: "Limit reached %{today}.",
```
yongjiajie, line 92: Since you already introduced a space in `%{today}`, we can do this:
```suggestion
    limitReachedRecent: "Limit reached %{time} ago%{today}.",
    limitReachedDate: "Limit reached on %{dateTime}%{today}.",
    limitReached: "Limit reached%{today}.",
```

yongjiajie (changes requested):

src/common/i18n/translations/en.ts
yongjiajie: @shermainelim might be good to run through Eida before changing this in the future, since it's something she introduced.

src/common/i18n/translations/en.ts
shermainelim: @ilhaoni Can I include a space between ago and %{today} ?

src/common/i18n/translations/en.ts
shermainelim: made a mistake here. Noted.

src/components/CustomerQuota/NoQuota/NoQuotaCard.test.ts
```diff
@@ -276,3 +277,125 @@ describe("NoQuotaCard utility functions", () => {
...
+      transactionTime: new Date("2020-10-21T08:04:28.038Z"),
+    },
+    {
+      category: "replacement-appreciation-gift",
+      quantity: 1,
+    },
+  ],
+};
+
+describe("checkifAppealAvailable", () => {
```
millasml, line 387: We usually name the test suite based on the function, so this should be "`checkHasAppealProduct`"

millasml: looking at the function, i think we should test at least 3 things:
1) where allProducts has no appeal category type
2) where all product has appeal category type, and quota quantity for this category > 0
3) where all product has appeal category type, and quota quantity for this category == 0

I think you covered 3 and 2, but you should add another test case for 1!

src/components/CustomerQuota/NoQuota/NoQuotaCard.test.ts
```diff
@@ -276,3 +277,125 @@ describe("NoQuotaCard utility functions", () => {
...
+  it("should not return appeal because product has been appealed", () => {
```
millasml, line 388: "should return false when quota for the appeal category is zero"

src/components/CustomerQuota/NoQuota/NoQuotaCard.test.ts
```diff
@@ -276,3 +277,125 @@ describe("NoQuotaCard utility functions", () => {
...
+    expect.assertions(1);
+    expect(checkHasAppealProduct(allProducts, allQuotaResponseNoAppeal)).toBe(
+      false
+    );
+  });
+
+  it("should return that has enough appeal quota to request for appeal", () => {
```
millasml, line 395: "should return true when quota for the appeal category is more than zero"

millasml (changes requested):

millasml (approved):

src/components/CustomerQuota/NoQuota/NoQuotaCard.test.ts
```diff
@@ -2,62 +2,258 @@ import {
...
-            value: "AAA987654321",
-          },
-        ],
-        transactionTime: latestTransactionTime,
-      },
+  describe("checkHasAppealProduct", () => {
+    const allQuotaResponseNoAppeal: Quota = {
+      globalQuota: [
+        {
+          category: "appreciation-gift",
```
yongjiajie, line 73: Let's not use something from the policy, would be good to use something generic like `gift` and `replacement-gift`.

Please help to change the rest!

yongjiajie (changes requested):
All looks good, just some minor changes to the mock data.

yongjiajie (approved):
Looks good!

src/components/CustomerQuota/NoQuota/NoQuotaCard.test.ts
```diff
@@ -2,62 +2,258 @@ import {
...
-        transactionTime: latestTransactionTime,
-      },
+  describe("checkHasAppealProduct", () => {
+    const allQuotaResponseNoAppeal: Quota = {
+      globalQuota: [
+        {
+          category: "gift",
+          quantity: 0,
+          quotaRefreshTime: 1634800083079,
+          transactionTime: new Date("2020-10-21T07:08:03.079Z"),
```
ilhaoni, line 76: haha small nit, shall we use latestTransactionTime? 
```suggestion
          transactionTime: latestTransactionTime,
```

src/components/CustomerQuota/NoQuota/NoQuotaCard.test.ts
```diff
@@ -2,62 +2,258 @@ import {
...
+        },
+        {
+          category: "replacement-gift",
+          quantity: 0,
+          quotaRefreshTime: 1634800102815,
+          transactionTime: new Date("2020-10-21T07:08:22.815Z"),
```
ilhaoni, line 82:
```suggestion
          transactionTime: latestTransactionTime,
```

ilhaoni (commented):
Reviewed! last small bits~ Hmm I think we can use the lastTransactionTime defined, rather than initialising a new date haha.

ilhaoni (approved):
LGTM! :D

PR #228, opened 2020-10-20
chore: enable new rules by prettier 2.0

szenius:
This PR re-enables the new defaults by prettier 2.0 and fixes all the issues due to the new defaults~

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf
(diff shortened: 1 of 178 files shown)


szenius:
I think the failure is due to what Seb mentioned

>endOfLine: Essentially last time was the wild west where files could have both types of line endings  (diff for windows and unix-based), now prettier wants to fix  them all to be LF (i.e. unix wins).

Going to check again!!

yongjiajie (approved):
Looks great, thanks for doing this!

sebastianquek (approved):
Thanks for updating this 🚀

PR #229, opened 2020-10-20
feat: reload translation on app state update to active

szenius:
[Story](https://www.notion.so/Chinese-Translation-Enhancements-eea2fb29a0a5414e9139f546a2b971f0)

This PR adds reloading of translation when locale changes, without needing the app to be restarted. This [only works for Android phones](https://docs.expo.io/versions/latest/sdk/localization/#localizationgetlocalizationasync).


sebastianquek (approved):
Looks good to me! Sorry for the delay!!

yongjiajie (approved):
Tried it out and looks great! Thanks for noticing this was possible on Android!

szenius:
@yongjiajie Seb's idea!!

PR #232, opened 2020-10-23
Feat/chinese translation for redemption stats

shermainelim:
Related to: Chinese translation for app Redemption Statistics

-call i18ntWithValidator directly for util.ts via Appeal as outside react component


src/hooks/useDailyStatistics/utils.test.ts
```diff
@@ -231,6 +232,12 @@ describe("countTotalTransactionsAndByCategory", () => {
 
   it("should have appeal alertDescription 'via appeal' if product is from an appeal flow", () => {
     expect.assertions(1);
+    console.log(
+      countTotalTransactionsAndByCategory(
+        pastTransactionsWithAppeal,
+        campaignPolicy
+      )
+    );
```
yongjiajie, line 240: Is this still necessary?

src/components/DailyStatistics/TitleStatistic.tsx
```diff
@@ -61,15 +62,19 @@ export const TitleStatisticComponent: FunctionComponent<TitleStatisticComponent>
...
+      </AppText>
       <AppText style={styles.statText}>{totalCount?.toLocaleString()}</AppText>
       <AppText style={styles.smallText}>
-        Last distributed at{" "}
+        {i18nt("redemptionStats", "lastDistributedAt")}{" "}
         {lastTransactionTime !== null
           ? format(lastTransactionTime, "h:mma")
-          : "-"}
+          : "-"}{" "}
+        {i18nt("redemptionStats", "distributedTime")}
```
yongjiajie, line 77: This `i18nt` copy can be moved to line 75, since it is possible to feed dynamic values into `i18nt` for interpolation. You can refer to existing implementations of `i18nt` for that.

This can be used to show different copies for `en` and `zh`, where en only has the `lastTransactionTime`, but `zh` has the `lastTransactionTime` and additional text.

src/components/DailyStatistics/TitleStatistic.tsx
```diff
@@ -61,15 +62,19 @@ export const TitleStatisticComponent: FunctionComponent<TitleStatisticComponent>
...
+        {i18nt("redemptionStats", "youDistributed")}
+      </AppText>
       <AppText style={styles.statText}>{totalCount?.toLocaleString()}</AppText>
       <AppText style={styles.smallText}>
-        Last distributed at{" "}
+        {i18nt("redemptionStats", "lastDistributedAt")}{" "}
         {lastTransactionTime !== null
           ? format(lastTransactionTime, "h:mma")
-          : "-"}
+          : "-"}{" "}
```
yongjiajie, line 76: I think we can incorporate the spaces into the `i18nt` copies, would be easier to manage the strings that way, and would result in less maintenance at the component level. Also, this would allow us flexibility in determining which languages will display a space, and which will not.

yongjiajie (changes requested):

src/components/DailyStatistics/TitleStatistic.tsx
```diff
@@ -61,15 +62,20 @@ export const TitleStatisticComponent: FunctionComponent<TitleStatisticComponent>
...
       <AppText style={styles.smallText}>
-        Last distributed at{" "}
+        {i18nt("redemptionStats", "lastDistributedAt")}
         {lastTransactionTime !== null
-          ? format(lastTransactionTime, "h:mma")
-          : "-"}
+          ? i18nt("redemptionStats", "distributedTime", undefined, {
+              distributionTime: format(lastTransactionTime, "h:mma"),
+            })
+          : i18nt("redemptionStats", "noDistributedTime")}
```
yongjiajie, line 78: You can extract this logic out and use a single `i18nt` call.

yongjiajie (changes requested):

src/common/i18n/translations/en.ts
```diff
@@ -1,6 +1,17 @@
 import { Translations } from "./type";
 
 export const en: Translations = {
+  redemptionStats: {
+    youDistributed: "You distributed",
+    lastDistributedAt: "Last distributed at ",
+    distributedTime: "%{distributionTime} ",
+
+    back: "Back",
```
yongjiajie, line 9: I personally don't see the value of separating the sentence. Let me know what you think.
I think we can just do this:

```suggestion
    lastDistributedAt: "Last distributed at %{lastTransactionTime}",
    back: "Back",
```

src/common/i18n/translations/zh.ts
```diff
@@ -1,6 +1,17 @@
 import { Translations } from "./type";
 
 export const zh: Translations = {
+  redemptionStats: {
+    youDistributed: "您已分发",
+    lastDistributedAt: "最后在 ",
+    distributedTime: "%{distributionTime} 分发",
+
+    back: "返回",
```
yongjiajie, line 9: Similar for here:

```suggestion
    lastDistributedAt: "最后在 %{lastTransactionTime} 分发",
    back: "返回",
```

src/components/DailyStatistics/TitleStatistic.tsx
```diff
@@ -61,15 +62,21 @@ export const TitleStatisticComponent: FunctionComponent<TitleStatisticComponent>
...
+        {i18nt("redemptionStats", "youDistributed")}
+      </AppText>
       <AppText style={styles.statText}>{totalCount?.toLocaleString()}</AppText>
       <AppText style={styles.smallText}>
-        Last distributed at{" "}
-        {lastTransactionTime !== null
-          ? format(lastTransactionTime, "h:mma")
-          : "-"}
+        {i18nt("redemptionStats", "lastDistributedAt")}
+        {i18nt("redemptionStats", "distributedTime", undefined, formatTime)}
```
yongjiajie, line 79: Similar for this, we can just use one sentence:

```suggestion
        {i18nt("redemptionStats", "lastDistributedAt", undefined, formattedLastTransactionTime)}
```

src/components/DailyStatistics/TitleStatistic.tsx
```diff
@@ -61,15 +62,21 @@ export const TitleStatisticComponent: FunctionComponent<TitleStatisticComponent>
   onPressPrevDay,
   onPressNextDay,
 }) => {
+  const { i18nt } = useTranslate();
+  const formatTime =
+    lastTransactionTime !== null
+      ? { distributionTime: format(lastTransactionTime, "h:mma") }
+      : { distributionTime: "-" };
+
```
yongjiajie, line 70: Since they're both using the same key, we can move the conditional in:

```suggestion
  const formattedLastTransactionTime =
      { lastTransactionTime: lastTransactionTime !== null ? format(lastTransactionTime, "h:mma") : "-" }

```

yongjiajie (changes requested):

yongjiajie (changes requested):
The product names, quantities, and description alert in `TransactionHistoryCard` should be translated as well. You can look at `ItemContent` for reference on how to do so.

src/components/DailyStatistics/TransactionHistoryCard.tsx
```diff
@@ -58,30 +64,36 @@ export const TransactionHistoryCardComponent: FunctionComponent<TransactionHisto
...
-                <AppText style={styles.categoryName}>{item.name}</AppText>
+                <AppText style={styles.categoryName}>
+                  {c13nt(item.name)}
+                </AppText>
                 <View style={styles.transactionText}>
-                  <AppText style={styles.quantityText}>
-                    {item.quantityText}
+                  <AppText style={styles.quantity}>{item.quantity}</AppText>
+                  <AppText style={styles.unit}>
+                    {c13ntForUnit(item.unit)?.label}
```
millasml, line 81: This assumes that all the unit labels are postfixes. Is that okay?

shermainelim: in CheckoutSuccessCard, is using "POSTFIX". Is that ok if we use that?

millasml: instead of doing this, I think you should pass it to `formatQuantityText`. remember to replace the label with the translated one

shermainelim: updated

src/components/DailyStatistics/StatisticsHeader.tsx
```diff
@@ -66,9 +67,14 @@ export const StatisticsHeaderComponent: FunctionComponent<StatisticsHeader> = ({
           size={size(4)}
           color={color("grey", 0)}
         />
-        <AppText style={styles.backText}>Back</AppText>
+        <AppText style={styles.backText}>
+          {" "}
```
sebastianquek, line 71: Do we need this?

src/hooks/useDailyStatistics/utils.ts
```diff
@@ -39,16 +41,19 @@ export const countTotalTransactionsAndByCategory = (
...
+
+          quantity: totalQuantityInCategory,
+          unit: findItemByCategory?.quantity.unit || {
+            type: "POSTFIX",
+            label: ` ${i18ntWithValidator(
+              "checkoutSuccessScreen",
+              "quantity"
+            )}`,
+          },
+
```
sebastianquek, line 53: Pardon me if I'm asking a question raised before, am also not 100% familiar with the context.

What the reason for replacing `formatQuantityText` here and letting the consumers handle it instead? I noticed the current code uses `quantityText` so the consumers of `useDailyStatistics` hook don't need to care about the specific details on how to format the quantity—they can just use the `quantityText` directly.

src/components/DailyStatistics/TransactionHistoryCard.tsx
```diff
@@ -25,7 +25,7 @@ const styles = StyleSheet.create({
     fontSize: fontSize(0),
     marginBottom: size(3),
   },
-  quantityText: {
+  unit: {
```
sebastianquek, line 28: This can be reverted to `quantityText`  since it's no more specifically styling the unit!

src/hooks/useDailyStatistics/useDailyStatistics.tsx
```diff
@@ -16,6 +19,9 @@ export type StatisticsHook = {
   }[];
   error?: Error;
   loading: boolean;
+  c13ntForUnit: (
+    unit: CampaignPolicy["quantity"]["unit"]
+  ) => CampaignPolicy["quantity"]["unit"];
```
sebastianquek, line 24: Why do we need to export this function for this particular hook?

shermainelim: for the testing. destructure c13ntForUnit and mock hocks.

src/components/DailyStatistics/TransactionHistoryCard.tsx
shermainelim: reverted

src/hooks/useDailyStatistics/useDailyStatistics.tsx
sebastianquek: For the testing, we don't need the c13ntForUnit to be actually instantiated, we can just mock it there. Notice that the utils is calling the main hook, which is weird because the test of utils do not need to have a direct dependency on `useDailyStatistics`

shermainelim: have removed the dependency and using mocks.

sebastianquek: Can we delete this export then?

shermainelim: Yes, deleted.

src/hooks/useDailyStatistics/utils.test.tsx
```diff
@@ -187,33 +198,37 @@ describe("countTotalTransactionsAndByCategory", () => {
...
+          quantityText: "999 qty",
           descriptionAlert: undefined,
           order: 2,
         },
       ],
     });
   });
 
-  it("should return category with category as name if transacted item is not in policies", () => {
+  it("should return category with category as name if transacted item is not in policies", async () => {
```
sebastianquek, line 224: Noticed you adde `async` to several tests, are they necessary?

src/hooks/useDailyStatistics/utils.test.tsx
```diff
@@ -155,24 +161,29 @@ describe("countTotalTransactionsAndByCategory", () => {
...
+        jest.fn()
+      )
     ).toStrictEqual({
       summarisedTotalCount: 4019,
       summarisedTransactionHistory: [
         {
           category: "instant-noodles",
           name: "🍜 Instant Noodles",
-          quantityText: "999 pack(s)",
+          quantityText: "999 qty",
```
sebastianquek, line 179: I noticed now we're just testing for `qty` while previously we explicitly tested for `pack(s)`. We should be mindful of changing test cases as there are some functionalities being tested that could be lost  with such changes.

For this particular case, is it useful to test for `pack(s)`?

shermainelim: added mock

sebastianquek:
Hmm, there are many changes that are not related to this branch, could you check again?

PR #238, opened 2020-11-10
fix: added scroll view for nav drawer

shermainelim:
This PR adds... 

Related to: [Bug: IOS & Android Scale Right Tab](https://www.notion.so/701364d4fbbf4ec1801b766c589973b9?v=24aa9675206d4da9a171ca4073d81b5b&p=9c15e648671e482ba2b53071cc0c2412)

- To scale at max font and max display size for IOS & Android.

**For IOS**

Max Font & Max Display Size:

iosmax

Default:
iosdefault

**Android**
Max Font & Max Display Size:

Note: Scaled Help & Support to appear within view so that users will know and can scroll down.

Screenshot 2020-11-18 at 9 35 04 AM

Alternative:
Screenshot 2020-11-18 at 9 35 25 AM

Default:
Screenshot 2020-11-18 at 9 33 25 AM

**Dynamic Scaling**

ScrollView is parent view, the parent view wouldn't allow more height so when specify with flexGrow, the container can grow up to flex 1. 
So when the View set to flex 1 , then it can take the space.

-using const { height: HEIGHT } = Dimensions.get("window"); 
- with minHeight: HEIGHT * 0.6  
- This will always be minimum at 60% of the screen height. 

If the phone is really small, and the screen is made bigger, then it will always have the space. If is a large screen, with justifyContent:"space-between", there will be space.

- Removed marginBottom so that the bottom section is adjusted to the bottom.

### DIFF
--- a/src/components/Layout/DrawerNavigation.tsx
+++ b/src/components/Layout/DrawerNavigation.tsx
@@ -164,7 +164,7 @@ export const DrawerNavigationComponent: FunctionComponent<DrawerContentComponent
             borderTopWidth: 1,
             flex: 1,
             justifyContent: "space-between",
-            minHeight: HEIGHT * 0.6,
+            minHeight: HEIGHT * 0.55,
           }}
         >
           <View>
--- a/src/components/Layout/DrawerNavigation.tsx
+++ b/src/components/Layout/DrawerNavigation.tsx
@@ -5,6 +5,7 @@ import {
   TouchableOpacity,
   View,
   ScrollView,
+  Platform,
 } from "react-native";
 import {
   DrawerContentComponentProps,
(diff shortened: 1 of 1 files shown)


sebastianquek:
I noticed for the default, it's now different from the existing design where the hyperlinks are flushed to the bottom of the screen. Can you make sure this designed is maintained?

shermainelim:
fixed.

sebastianquek:
I've updated the code with the approach I previously mentioned. You were adding the styles to the wrong `View` and hence you didn't see any result.

Notice that in the case below, the hardcoded `0.55*HEIGHT` approach doesn't work too well as the version is hidden and there's too much unused whitespace. In general, we should always strive for ways that don't require hardcoding of specific values.

| Before | After |
| ---- | ---- |
| simulator_screenshot_D0ED9863-103B-408B-A150-F2276E2F1E86 | simulator_screenshot_FE6E5AD0-2895-4362-BB0E-609163ECB127 |

sebastianquek (approved):

PR #254, opened 2020-12-07
chore: CI changelog improvements

sebastianquek:
Currently trying to rerun a failed `create release` github action fails as the changelog action errors out. By setting the explicit `head-ref` and `base-ref`, we let the action know the exact changelog to generate, and update the old one if need be. This allows us to rerun the failed actions multiple times.


dalsontws (approved):
thanks so much for this, LGTM!

PR #255, opened 2020-12-08
chore: ensure tags are fetched properly

sebastianquek:
Tags were not being fetched properly in `create_release`, so `git tag` was returning only the latest tag, causing the changelog generation to fail.


ilhaoni (approved):
LGTM! :D

PR #263, opened 2020-12-14
fix: Multiple Campaign ID Validation 

yongjiajie:
[Notion link](https://www.notion.so/When-I-hv-multiple-distributions-campaigns-logged-in-switching-from-one-to-another-causes-validatio-7f7cf92215b6496baec0e1c56a818407) 

This PR adds... 

- Fix that ensures the correct ID validation rules are loaded when switching between different campaigns.

### Why did this bug occur?

The condition used to determine if the existing validation rules should be used was always returning `true` since it was matching `undefined === undefined`. This resulted in the bug where the first loaded campaign's validation rules would always be used, instead of using each campaign's validation rules when changing between campaigns.

### DIFF
--- a/src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
+++ b/src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
@@ -169,9 +169,11 @@ const CollectCustomerDetailsScreen: FunctionComponent<NavigationFocusInjectedPro
     // in the event the saved selection not found.. will always fall back to the first idType in array
     setSelectedIdType(
       selectionArray.some((selection) => {
-        return selection.label && selectedIdType.label
-          ? selection.label === selectedIdType.label
-          : false;
+        return (
+          selection.label &&
+          selectedIdType.label &&
+          selection.label === selectedIdType.label
+        );
       })
         ? selectedIdType
         : selectionArray[0]
--- a/src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
+++ b/src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
@@ -168,9 +168,11 @@ const CollectCustomerDetailsScreen: FunctionComponent<NavigationFocusInjectedPro
   useEffect(() => {
     // in the event the saved selection not found.. will always fall back to the first idType in array
     setSelectedIdType(
-      selectionArray.some(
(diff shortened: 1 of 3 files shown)


src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
```diff
@@ -168,9 +168,11 @@ const CollectCustomerDetailsScreen: FunctionComponent<NavigationFocusInjectedPro
   useEffect(() => {
     // in the event the saved selection not found.. will always fall back to the first idType in array
     setSelectedIdType(
-      selectionArray.some(
-        (selection) => selection.label === selectedIdType.label
-      )
+      selectionArray.some((selection) => {
+        return selection.label && selectedIdType.label
+          ? selection.label === selectedIdType.label
+          : false;
```
sebastianquek, line 174: Can we do `selection.label && selectedIdType.label && selection.label === selectedIdType.label` instead?

sebastianquek:
Looks like the key issue is that `selectedIdType` is not cleared on changing of campaign, hence the component mounts and uses a stale context. This fix will work, but concept of stale context may come back to haunt us.

src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
yongjiajie: This is done!

yongjiajie:
> Looks like the key issue is that `selectedIdType` is not cleared on changing of campaign, hence the component mounts and uses a stale context. This fix will work, but concept of stale context may come back to haunt us.

I have added changes to ensure that `selectedIdType` is cleared on changing of campaign. Please help to check if my implementation is sound!

src/context/identification.tsx
```diff
@@ -26,9 +28,12 @@ export const IdentificationContextProvider: FunctionComponent = ({
     defaultSelectedIdType
   );
 
+  const resetSelectedIdType = (): void =>
+    setSelectedIdType(defaultSelectedIdType);
+
```
sebastianquek, line 33: May want to wrap this in a `useCallback` so that whenever its ancestors updates, this function will have the same reference. This helps with ensuring stable references for `useEffect` dependencies

dalsontws (approved):
tested and LGTM!🚀

sebastianquek (approved):
LGTM!

PR #282, opened 2021-01-18
fix: remove phone number length restriction

sebastianquek:
This PR removes the formatting of phone numbers which go above the `exampleNumber`'s length.


yongjiajie (approved):
LGTM! Thanks for helping to do this! 🎉

PR #285, opened 2021-01-22
fix: issue with mergeWithCart using the quotaResponse incorrectly

sebastianquek:
The quota endpoint returns an identifierInputs field that represents the values of the latest transaction. In `mergeWithCart` we were treating those values as the template for the identifierInputs. This meant that we were incorrectly using the latest transaction values as the initialised value for the cart.

This works for most cases since the identifier inputs returned from the last transaction will match the identifier inputs from the latest transaction. However, the assumption isn't always valid and this PR corrects this logic.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -190,7 +190,6 @@ const ItemQuota = t.intersection([
   t.partial({
     quotaRefreshTime: t.number,
     transactionTime: DateFromNumber,
-    identifierInputs: t.array(IdentifierInput),
   }),
 ]);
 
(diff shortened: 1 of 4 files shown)


src/types.ts
```diff
@@ -190,7 +190,6 @@ const ItemQuota = t.intersection([
   t.partial({
     quotaRefreshTime: t.number,
     transactionTime: DateFromNumber,
-    identifierInputs: t.array(IdentifierInput),
```
szenius, line 193: Is there a corresponding backend PR to remove this from quota response?

sebastianquek: Not at the moment, will add a separate task for it. Am thinking that it's not as high priority since it doesn't need to be in place for the return flow to work.

szenius: As long as we are tracking it somewhere, else it might cause future confusion :)

szenius (approved):
Approved with understanding that there will be a task for backend 🙏

PR #291, opened 2021-01-27
chore: allow manual e2e workflow runs

sebastianquek:
GitHub actions now support manually triggered workflows. Have enabled this for e2e tests so that we can run it on the fly.
https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_dispatch

Example:
image

### DIFF
--- a/.github/workflows/test_e2e.yml
+++ b/.github/workflows/test_e2e.yml
@@ -12,6 +12,8 @@ on:
       - master
     types:
       - completed
+  # Allow manual runs
+  workflow_dispatch:
 
 jobs:
   e2e_test:


.github/workflows/test_e2e.yml
```diff
@@ -12,6 +12,8 @@ on:
       - master
     types:
       - completed
+  # Allow manual runs
+  workflow_dispatch:
```
ilhaoni, line 16: hmm, i think may need the inputs field here, if we're pasting a url in? :O 

or currently its more to trigger it when needed~ xD

sebastianquek: trigger when needed, so no url needed, check out the code here: https://github.com/sebastianquek/github-actions-test/blob/master/.github/workflows/test_env_var.yml

```
on:
  push:
  pull_request:
  workflow_dispatch:
```

ilhaoni: okk! got it haha~ thanks~

ilhaoni (approved):
LGTM! :D

PR #293, opened 2021-01-29
feat: phone number scanning

sebastianquek:
[Notion link](https://www.notion.so/Phone-number-scanning-68b12284ee2e40f9bf4fdc37e00b07ae) 

This PR adds ability to scan a phone number.
* Fixes alignment issue when the scan button is enabled
* Add logic to handle the scanning of the following types of numbers:
  * `91234567`: no country code, will default to SG
  * `+65 81234567`, `+12024561414`: has country code, will update the country code input to match what was scanned
  * `65 91234567`, `6581234567`, `12024561414`: has country code without `+` sign, will default to SG.

### DIFF
--- a/src/components/CustomerQuota/ItemsSelection/IdentifierLayout/IdentifierPhoneNumberInput.tsx
+++ b/src/components/CustomerQuota/ItemsSelection/IdentifierLayout/IdentifierPhoneNumberInput.tsx
@@ -1,29 +1,61 @@
 import React, { FunctionComponent, useState, useEffect } from "react";
-import { createFullNumber } from "../../../../utils/validatePhoneNumbers";
+import {
+  createFullNumber,
+  parsePhoneNumber,
+} from "../../../../utils/validatePhoneNumbers";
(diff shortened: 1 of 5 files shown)


andrehadianto (approved):
tested, LGTM ✈️ 

just curious, why should the country code of the input number be automatically removed as the user enters the number? Feels like it's going to catch them by surprise.

sebastianquek:
> just curious, why should the country code of the input number be automatically removed as the user enters the number? Feels like it's going to catch them by surprise.

Good catch, this is actually to help users who enter the country code in the input number field. Note that the country code will only be removed if the library detects that there's a valid phone number for that country. E.g.
1. User keys in "6588 8888" <- this is a valid SG landline number
2. User continues to key in 2 more eights: "6588 8888 88" <- In this case the lib will detect that given `countryCode = 65` and `phoneNumber = "6588 8888 88"`, the parsed phone number in international format is "+65 8888 8888", so the input field will show "8888 8888".

PR #300, opened 2021-02-03
feat: passport scanner fe

andrehadianto:
[Notion link](https://www.notion.so/Building-a-software-that-translate-information-from-Passport-Scanner-into-a-readable-QR-code-for-Sal-09941e155354434685394cb26731dbba) 

This PR adds... 
- Refactor ManualPassportInput to PassportInputSection
- scanner component on Passport Input section
- logic to check whether for passport input `hasScanner` which is checked from features, whether `alternateIds` has scannerType === "QR" or "NONE"
- logic to switch barcode type when selectedIdType === "passport" or not

- [x] I've added documentation in README/[Notion](https://www.notion.so/82b92fb1007640328dab9582c0a3694e?v=3b6ce48202cf40ad8450553799b13146)


src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
```diff
@@ -204,6 +204,11 @@ const CollectCustomerDetailsScreen: FunctionComponent<NavigationFocusInjectedPro
   const onBarCodeScanned: BarCodeScannedCallback = (event) => {
     if (isFocused && isScanningEnabled && event.data) {
       onCheck(event.data);
```
yongjiajie, line 206: Shouldn't this be removed? In the case where the type is not `Code39`, we won't be able to perform `onCheck` with the `passportId`, if this is still present.

```suggestion
```

src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
```diff
@@ -204,6 +204,11 @@ const CollectCustomerDetailsScreen: FunctionComponent<NavigationFocusInjectedPro
...
+      if (event.type.includes("Code39")) {
```
yongjiajie, line 207: May be better to capitalise this and then perform a comparison, so even when the capitalisation in `type` changes, you are still able to compare them.

src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
```diff
@@ -204,6 +204,11 @@ const CollectCustomerDetailsScreen: FunctionComponent<NavigationFocusInjectedPro
...
+        onCheck(event.data);
+      } else {
+        onCheck(JSON.parse(event.data).passportId);
```
yongjiajie, line 210: Not sure if this looks better, and you can easily access properties you add to data in the future too!

```suggestion
        const { passportId } = JSON.parse(data);
        onCheck(passportId);
```

src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
```diff
@@ -219,7 +224,17 @@ const CollectCustomerDetailsScreen: FunctionComponent<NavigationFocusInjectedPro
   const getInputComponent = (): JSX.Element => {
     return selectedIdType.label === "Passport" &&
       selectedIdType.scannerType === "NONE" ? (
-      <ManualPassportInput
+      <PassportInputSection
```
yongjiajie, line 227: Can change this to be `InputPassportSection` to be similar to `InputIdSection` 🤔

src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
```diff
@@ -219,7 +224,17 @@ const CollectCustomerDetailsScreen: FunctionComponent<NavigationFocusInjectedPro
...
+        submitId={() => onCheck(idInput)}
+      />
+    ) : selectedIdType.label === "Passport" &&
+      selectedIdType.scannerType === "QR" ? (
+      <PassportInputSection
+        hasScanner={true}
+        openCamera={() => setShouldShowCamera(true)}
         setIdInput={setIdInput}
         submitId={() => onCheck(idInput)}
       />
```
yongjiajie, line 240: What do you think about passing `selectedIdType.scannerType` down to the component instead? You can remove the duplication of `PassportInputSection` here by doing that.

```suggestion
      <PassportInputSection
        scannerType={selectedIdType.scannerType}
        openCamera={() => setShouldShowCamera(true)}
        setIdInput={setIdInput}
        submitId={() => onCheck(idInput)}
      />
```

src/components/CustomerDetails/PassportInputSection.tsx
```diff
@@ -0,0 +1,130 @@
...
+            fullWidth={true}
+            text={i18nt("collectCustomerDetailsScreen", "scanIdentification")}
+            icon={
+              <Feather
+                name="maximize"
+                size={size(2)}
+                color={color("grey", 0)}
+              />
+            }
+            onPress={openCamera}
```
yongjiajie, line 74: Can toggle between `openCamera` and `() => void` based on `scannerType` here!

src/components/CustomerDetails/PassportInputSection.tsx
```diff
@@ -0,0 +1,130 @@
...
+  };
+
+  useEffect(() => {
+    selectedCountry && passportNum
+      ? setIdInput(`${selectedCountry?.id}-${passportNum}`)
+      : setIdInput("");
+  }, [selectedCountry, passportNum, setIdInput]);
+
+  const getScannerComponent = (): JSX.Element | null => {
+    return hasScanner ? (
```
yongjiajie, line 61: Can replace `hasScanner` with check for `scannerType`!

src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
```diff
@@ -238,6 +253,14 @@ const CollectCustomerDetailsScreen: FunctionComponent<NavigationFocusInjectedPro
     navigation.navigate("DailyStatisticsScreen");
   };
 
+  const getBarcodeType = (): any[] => {
+    if (selectedIdType.scannerType === "QR")
+      return [BarCodeScanner.Constants.BarCodeType.qr];
+    else if (selectedIdType.scannerType === "CODE_39")
+      return [BarCodeScanner.Constants.BarCodeType.code39];
+    else return [null];
```
yongjiajie, line 261: This removes the notion that there is a default of `code39`, which was true before. Might be better to keep the original logic as a fail-safe!

Let me know what you think!

```suggestion
    if (selectedIdType.scannerType === "QR") {
      return [BarCodeScanner.Constants.BarCodeType.qr];
    } else if (selectedIdType.scannerType === "CODE_39") {
      return [BarCodeScanner.Constants.BarCodeType.code39];
    } else {
      return features?.id.scannerType === "QR"
              ? [BarCodeScanner.Constants.BarCodeType.qr]
              : [BarCodeScanner.Constants.BarCodeType.code39];
    }
```

yongjiajie (changes requested):
Just some small changes! 💪

src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
andrehadianto: oh shoot. thankss

src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
andrehadianto: this one is because `event` looks like this
```
Object {
  "data": "S0000001I",
  "target": 689,
  "type": "org.iso.Code39",
}
```
should I still change it though? wdyt 🤔

src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
andrehadianto: good idea!

src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
yongjiajie: How about this?

```suggestion
      if (event.type === BarCodeScanner.Constants.BarCodeType.code39)) {
```

Alternatively, do you think we should:

- Pass `passportId` if present in `event.data`
- Pass `event.data` otherwise

Instead of changing what we check based on the scanner type.

andrehadianto: i think we should pass `passportId`, cus as of the story, QR for passport is an `object` while the Barcode for NRIC is a `string`, reason being we want to use `object` so that we can pass other data if needed.
With that said, I think can use the provided constant instead as you suggested!

yongjiajie (approved):
🚀 🌕

PR #301, opened 2021-02-08
feat: only show payment receipt field upon chargeable

ilhaoni:
[Notion link](https://www.notion.so/Update-policy-to-charge-lost-TT-Tokens-if-it-s-lost-for-a-2nd-time-Enter-Payment-receipt-number-28a3cb5dfdbc4e9e95a21a1b1490db0a) 

This PR adds the payment receipt field in IdentifierInputs if it's chargeable. Relies on backend api PR532. WIP!

Will be using filter instead of splice, mainly because `splice` makes changes to the `policies` 
```
Array.splice - will change the Array itself. (use: myArray.splice)
Array.filter - will return the filtered Array. (use: myFilteredArray = Array.filter)
```

### DIFF
--- a/src/common/i18n/translations/type.ts
+++ b/src/common/i18n/translations/type.ts
@@ -281,7 +281,6 @@ export type Translations = {
       title: string;
       body?: string;
       primaryActionText: string;
-      secondaryActionText?: string;
     };
     wrongFormatCountryCode: {
       title: string;
--- a/src/common/i18n/translations/type.ts
+++ b/src/common/i18n/translations/type.ts
@@ -277,6 +277,12 @@ export type Translations = {
       primaryActionText: string;
       secondaryActionText?: string;
     };
+    wrongPaymentReceiptNumber: {
+      title: string;
+      body?: string;
+      primaryActionText: string;
+      secondaryActionText?: string;
+    };
     wrongFormatCountryCode: {
       title: string;
       body?: string;
--- a/src/components/CustomerQuota/ItemsSelection/Item.tsx
+++ b/src/components/CustomerQuota/ItemsSelection/Item.tsx
@@ -28,8 +28,8 @@ export const Item: FunctionComponent<{
     item.label.includes("receipt number")
   );
 
-  // let newIdentifiers = identifiers;
-  // const newCartItem = cartItem;
+  const newIdentifiers = identifiers;
+  const newCartItem = cartItem;
 
   // these idenfitifiers should only be shown if this redemption is chargeable
   // so we remove them if otherwise
@@ -47,13 +47,8 @@ export const Item: FunctionComponent<{
     //   (identifier) => identifier.label != "Payment receipt number"
     // );
 
-    const removedFromCart = cartItem.identifierInputs.splice(
-      indexOfReceiptField,
-      1
-    );
-    const removedFromIden = identifiers.splice(indexOfReceiptField, 1);
-    console.log(removedFromCart);
-    console.log(removedFromIden);
+    newCartItem.identifierInputs.splice(indexOfReceiptField, 1);
+    newIdentifiers.splice(indexOfReceiptField, 1);
   }
 
   return (
@@ -77,9 +72,9 @@ export const Item: FunctionComponent<{
       )}
       {cartItem.maxQuantity > 0 && identifiers.length > 0 && (
         <ItemIdentifiersCard
-          cartItem={cartItem}
+          cartItem={newCartItem}
           updateCart={updateCart}
-          identifiers={identifiers}
+          identifiers={newIdentifiers}
         />
       )}
     </View>
--- a/src/components/CustomerQuota/ItemsSelection/Item.tsx
+++ b/src/components/CustomerQuota/ItemsSelection/Item.tsx
@@ -45,8 +45,8 @@ const isPodChargeable = (
  * Shape of identifiers and cartItem.identifierInputs are slightly different and
  * hence require to filter separately
  *
- * @param newIdentifiers identifiers that can be modified
- * @param newCartItem cartItem containing identifiers that can be modified
+ * @param identifiers identifiers that can be modified
+ * @param cartItem cartItem containing identifiers that can be modified
  */
 const removePaymentReceiptField = (
   identifiers: PolicyIdentifier[],
--- a/src/components/CustomerQuota/ItemsSelection/Item.tsx
+++ b/src/components/CustomerQuota/ItemsSelection/Item.tsx
@@ -24,23 +24,36 @@ export const Item: FunctionComponent<{
   const { getProduct } = useContext(ProductContext);
   const identifiers = getProduct(cartItem.category)?.identifiers || [];
 
-  let newIdentifiers = identifiers;
-  const newCartItem = cartItem;
+  const indexOfReceiptField = identifiers.findIndex((item) =>
+    item.label.includes("receipt number")
+  );
 
+  // let newIdentifiers = identifiers;
+  // const newCartItem = cartItem;
+
+  // these idenfitifiers should only be shown if this redemption is chargeable
+  // so we remove them if otherwise
   if (
     identifiers.length > 0 &&
     cartItem.category.includes("lost") &&
-    cartItem.descriptionAlert !== "*chargeable"
+    cartItem.descriptionAlert !== "*chargeable" &&
+    indexOfReceiptField != -1
   ) {
     console.log("test");
-    console.log(cartItem.category);
-    console.log(identifiers);
-    newIdentifiers = identifiers.filter(
-      (identifier) => identifier.label != "Payment receipt number"
+    // newCartItem.identifierInputs = newCartItem.identifierInputs.filter(
+    //   (identifier) => identifier.label != "Payment receipt number"
+    // );
+    // newIdentifiers = identifiers.filter(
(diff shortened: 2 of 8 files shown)


andrehadianto (changes requested):

src/common/i18n/translations/type.ts
```diff
@@ -272,6 +272,12 @@ export type Translations = {
       primaryActionText: string;
       secondaryActionText?: string;
     };
+    wrongPaymentReceiptNumber: {
+      title: string;
+      body?: string;
+      primaryActionText: string;
+      secondaryActionText?: string;
```
szenius, line 279: Actually no need add this, since no `secondaryActionText`. I know some of the other objects in here have it but I intended to clear those out

src/components/CustomerQuota/ItemsSelection/Item.tsx
```diff
@@ -22,6 +22,24 @@ export const Item: FunctionComponent<{
...
+    cartItem.category.includes("lost") &&
+    cartItem.descriptionAlert !== "*chargeable"
+  ) {
+    newIdentifiers = identifiers.filter(
+      (identifier) => identifier.label != "Payment receipt number"
+    );
+    newCartItem.identifierInputs = newCartItem.identifierInputs.filter(
+      (identifier) => identifier.label != "Payment receipt number"
+    );
+  }
```
szenius, line 41: I'm conscious about this block being super specific to a campaign which is not supposed to be the design of the mobile app. So I'd like to know: did we discuss other alternatives, and why did we choose this solution?

If this is the sole/best solution, perhaps we can make it clearer what this chunk is meant for? E.g. wrap this up in a condition for "if this campaign is TT", or extracting this logic out into a helper named specifically for TT. In that case we at least know which campaign caused us to need this logic, and it'll be easier to remove this logic once it's no longer necessary.

src/utils/validateIdentifierInputs.ts
```diff
@@ -17,8 +17,17 @@ const isUniqueList = (list: string[]): boolean =>
...
-  for (const { value, validationRegex, textInputType } of identifierInputs) {
+  for (const {
+    label,
+    value,
+    validationRegex,
+    textInputType,
+  } of identifierInputs) {
     //TODO: switch between different errors based on campaign config
+    // TODO: can set the validation regex here once it's confirmed
+    if (label === "Payment receipt number" && !value) {
```
szenius, line 28: Could this be a good chance to make use of c13n for error messages? No need to do the migration all at once, just for this could be a good start

src/components/CustomerQuota/ItemsSelection/Item.tsx
```diff
@@ -22,6 +22,24 @@ export const Item: FunctionComponent<{
   const { getProduct } = useContext(ProductContext);
   const identifiers = getProduct(cartItem.category)?.identifiers || [];
 
+  let newIdentifiers = identifiers;
+  const newCartItem = cartItem;
```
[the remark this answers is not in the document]
dalsontws: previously we tried to update the `cartItem` and `identifiers` by using `splice`, but somehow the edit resulted in removing the receipt field permanently, hence we decided to use `filter` and create a new array.

src/components/CustomerQuota/ItemsSelection/Item.tsx
```diff
@@ -22,6 +22,24 @@ export const Item: FunctionComponent<{
...
+
+  // these identifiers should only be shown if this redemption is chargeable
+  // so we will filter them out if otherwise
+  if (
+    identifiers.length > 0 &&
+    cartItem.category.includes("lost") &&
```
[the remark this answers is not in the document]
dalsontws: agreed, making this change!

src/components/CustomerQuota/ItemsSelection/Item.tsx
dalsontws: currently, the mobile app doesnt conditionally render any identifiers, hence we were discussing on how we can best cater for this case, without making too much changes. one way we thought of was conditionally excluding specific identifiers (for this case, the receipt field). this will allow us to add all related additional identifiers in policies.

another possible solution was to conditionally include the receipt field when it's chargeable, but felt like perhaps including part of the policies in code might not be the way moving forward. 

i think we can look into extracting this logic for better clarity, but what do you think of the current solution?😛
(cc @ilhaoni)

szenius: >i think we can look into extracting this logic for better clarity, but what do you think of the current solution?😛

Actually what I am saying is if we have no other alternatives, we might want to extract this out into an independent block, so that it's easier to do cleanup next time too. e.g. extracting out into a function, and doing a `if` around it (perhaps checking for the campaign name or some kind of campaign identifier), with some comment above to explain the context for which this was added (which will hint at when we can remove it too).

dalsontws: good point! i've extracted the entire logic and added a check for tt campaign as well, lemme know if this looks ok!

src/components/CustomerQuota/ItemsSelection/Item.tsx
```diff
@@ -15,6 +16,56 @@ const styles = StyleSheet.create({
...
+ * Filters out the payment receipt identifier if not required
+ * which happens when pod is not chargeable
+ * Shape of identifiers and cartItem.identifierInputs are slightly different and
+ * hence require to filter separately
+ *
+ * @param newIdentifiers identifiers that can be modified
+ * @param newCartItem cartItem containing identifiers that can be modified
+ */
+const removePaymentReceiptField = (
+  newIdentifiers: PolicyIdentifier[],
```
andrehadianto, line 52: nitty-gritty but wdyt of the params to be `identifiers` and `cartItem` instead? 🤔

src/utils/validateIdentifierInputs.ts
dalsontws: discussed with @andrehadianto on this, further discussion with the FE tech debt team on the approach of c13n for campaign specific error messages will be required before making changes here! so will leave this as is!

src/components/CustomerQuota/ItemsSelection/Item.tsx
dalsontws: good point, made the change! lemme know its this looks ok

src/components/CustomerQuota/ItemsSelection/Item.tsx
```diff
@@ -15,6 +16,56 @@ const styles = StyleSheet.create({
   },
 });
 
+/**
+ * Check if the campaign contains tt-token
+ * to indicate pod camapaign
```
andrehadianto, line 21: typo hahhaa

src/utils/validateIdentifierInputs.test.ts
```diff
@@ -189,4 +189,13 @@ describe("validateIdentifierInputs", () => {
       ])
     ).toThrow("Enter or scan a different code.");
   });
+
+  it("should throw the specific error if the identifierInput is payment receipt number", () => {
```
andrehadianto, line 193: maybe the test case can be named more specifically? Like in this case, `if the identifierInput value is empty`. 

Also, will other types of "STRING" like `spaces` or `emojis` be validated?

andrehadianto (changes requested):

src/utils/validateIdentifierInputs.ts
szenius: I see so is the `TODO` on line 26 needed? Since it's part of the tech debt and is already tracked somewhere.

Also what is line 27 `TODO` for?

szenius: Validation regex is set in SSM policies right?

src/utils/validateIdentifierInputs.test.ts
dalsontws: yeap have added a new type, `PAYMENT_RECEIPT` that uses regex validation and added more tests!

src/utils/validateIdentifierInputs.ts
dalsontws: yup! currently default is set to alphanumeric, less than 20, will confirm on this and set in SSM policies.
also removed the TODOs

szenius (approved):
Cool cool cool 
image

src/components/CustomerQuota/ItemsSelection/Item.tsx
```diff
@@ -15,6 +16,56 @@ const styles = StyleSheet.create({
...
+  cartItem.category.includes("tt-token-lost") &&
+  cartItem.descriptionAlert === "*chargeable";
+
+/**
+ * Filters out the payment receipt identifier if not required
+ * which happens when pod is not chargeable
+ * Shape of identifiers and cartItem.identifierInputs are slightly different and
+ * hence require to filter separately
+ *
+ * @param newIdentifiers identifiers that can be modified
```
andrehadianto, line 48: change the `@param` to `identifiers` and `cartItem`!

andrehadianto (changes requested):

src/components/CustomerQuota/ItemsSelection/Item.tsx
ilhaoni: done!

andrehadianto (approved):
LGTM!! 🎉

PR #303, opened 2021-02-09
feat: add filtering by categories for past transactions

yongjiajie:
[Notion link](https://www.notion.so/After-return-TT-Token-non-chargeable-transaction-history-are-removed-from-view-but-retains-Chargea-6ea0b27743e7433c96432fb373b3d3cb) 

This PR adds... 

- added hooks for updating and resetting categories
- altered API call to transaction history endpoint

### DIFF
--- a/src/hooks/usePastTransaction/usePastTransaction.tsx
+++ b/src/hooks/usePastTransaction/usePastTransaction.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState, useContext } from "react";
+import { useEffect, useState, useContext, useCallback } from "react";
 import { PastTransactionsResult } from "../../types";
 import { usePrevious } from "../usePrevious";
 import {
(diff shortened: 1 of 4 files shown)


src/hooks/usePastTransaction/usePastTransaction.tsx
```diff
@@ -27,6 +29,25 @@ export const usePastTransaction = (
   const [error, setError] = useState<PastTransactionError | null>(null);
   const prevIds = usePrevious(ids);
   const { selectedIdType } = useContext(IdentificationContext);
+  const [categories, setCategories] = useState<string[]>();
```
sebastianquek, line 32: I noticed that the effect to run `fetchPastTransactions` depends on `categories`, what if we just pass in `categories` as a prop to `usePastTransaction`? So instead of imperatively updating the `categories` state, we can pass it in declaratively.

sebastianquek (commented):
Do add tests for these too (:

sebastianquek (approved):
Thanks for making the changes!!

PR #306, opened 2021-02-11
Feat: update quota screen appeals logic

andrehadianto:
[Notion link](https://www.notion.so/Translation-c13n-and-i18n-standardisation-a249a80094c946d490ba479f6ed58fae) 

This PR adds... 
- implement logic to fetch pastTransaction for the appeals dropdown


src/components/CustomerQuota/ItemsSelection/ItemStepper.tsx
```diff
@@ -14,6 +14,8 @@ import { ProductContext } from "../../../context/products";
 import { sharedStyles } from "./sharedStyles";
 
 export const ItemStepper: FunctionComponent<{
```
sebastianquek, line 16: How are you thinking of handling the addons for this component?

src/components/CustomerQuota/ItemsSelection/AddonsItems.tsx
```diff
@@ -1,13 +1,72 @@
...
+  sortTransactionsByCategory,
+} from "../utils";
 
-export const AddonsItems: FunctionComponent = () => {
-  return (
+const BIG_NUMBER = 99999;
+
+export const AddonsItems: FunctionComponent<{
+  ids: string[];
+  isShowAddons: boolean;
```
sebastianquek, line 19: I think something like `isShowAddonItems` will be better since now it sounds like whether to show the entire addons block or not.

src/components/CustomerQuota/ItemsSelection/AddonsItems.tsx
```diff
@@ -1,13 +1,72 @@
...
+}> = ({ ids, isShowAddons }) => {
+  const { sessionToken, endpoint } = useContext(AuthContext);
+  const { policies: allProducts } = useContext(CampaignConfigContext);
+  const { pastTransactionsResult, loading, error } = usePastTransaction(
```
sebastianquek, line 23: Can we do the following to avoid the need for line 29?
```suggestion
  const { pastTransactionsResult: sortedTransactions, loading, error } = usePastTransaction(
```

src/components/CustomerQuota/ItemsSelection/ItemStepper.tsx
andrehadianto: will remove them for now since the props are not compulsory.

For the handling, thinking of making them look similar to ItemCheckbox's one on another PR once this one is approved!

src/components/CustomerQuota/ItemsSelection/ItemCheckbox.tsx
```diff
@@ -32,7 +34,15 @@ export const ItemCheckbox: FunctionComponent<{
           showAddons={isShowAddonsItems}
         />
       }
-      addons={<AddonsItems />}
+      addons={
+        isChargeable ? (
+          <AddonsItems
+            ids={ids}
+            isShowAddonItems={isShowAddonsItems}
+            categoryFilter={[category]}
```
sebastianquek, line 42: Given that this will always be 1 item, do we need a list?

src/components/CustomerQuota/ItemsSelection/AddonsItems.tsx
```diff
@@ -1,13 +1,72 @@
...
+  const transactionsByCategoryMap = groupTransactionsByCategory(
+    sortedTransactions,
+    allProducts || [],
+    latestTransactionTime,
+    translationProps
+  );
+
+  const transactionsByCategoryList = sortTransactionsByCategory(
+    transactionsByCategoryMap
+  );
```
sebastianquek, line 51: Noticed that we're performing all these on every render, can you check the number of times this is being called? It may be better to use a `useEffect` such that it only updates when `sortedTransactions` updates

src/components/CustomerQuota/ItemsSelection/AddonsItems.tsx
```diff
@@ -1,13 +1,72 @@
...
+  return loading && isShowAddonItems ? (
```
sebastianquek, line 52: If `loading = false` and `isShowAddonItems = false`, then we'll show the list of transactions. Not sure if this is expected, what do you think?

src/components/CustomerQuota/TransactionsGroup.tsx
```diff
@@ -50,7 +50,7 @@ export interface Transaction {
 }
 
 export interface TransactionsGroup {
-  header: string;
+  header: string | undefined;
```
sebastianquek, line 53: Could you explain when the header will be undefined?

src/components/CustomerQuota/ItemsSelection/AddonsItems.tsx
andrehadianto: however the `maxTransactionToDisplay` will result in 0, so it will not render any transaction. My idea was so that the transactions are preloaded such that there won't be additional fetch

src/components/CustomerQuota/TransactionsGroup.tsx
andrehadianto: based on the eida's UI, she wants the list to show `transactionsByCategoryList` without showing the category header (in this case is `Lost Token`). To handle this, I added `header={undefined}` to remove the header in `AddonsItems`. What do you thinkk?

src/components/CustomerQuota/ItemsSelection/AddonsItems.tsx
sebastianquek: Yep so there are 2 points of logic that affect this conditional render, `isShowAddonItems` and also `maxTransactionsToDisplay`. This can be confusing to reason about. Note that `maxTransactionsToDisplay` is meant as a way to limit the number of transactions on screen, less of a boolean to control if any transactions should be shown.
Also, given that the rendering of `AddonsItems` will already fetch the past transactions, the fetch will always take place.

What if we separate out the logic of whether `loading` and whether `isShowAddonItems` is true? i.e. if loading, regardless of isShowAddonItems, show the activity indicator. If not loading, then check for `isShowAddonItems`

src/components/CustomerQuota/ItemsSelection/ItemCheckbox.tsx
andrehadianto: this one is from @yongjiajie 's pr. I think he meant to make it generic so that we can filter multiple categories if needed. What do you think?

sebastianquek: Yea the API remains, just that on this particular component it should only have 1 category

andrehadianto: oh ok, understood your comment

src/components/CustomerQuota/ItemsSelection/AddonsItems.tsx
```diff
@@ -1,17 +1,92 @@
...
+                {...transactionsByCategory}
+                header={undefined}
+              />
+            )
+          )
+        : undefined}
+    </>
   ) : (
-    <View></View>
+    <></>
```
sebastianquek, line 90: Can just return `null`

src/components/CustomerQuota/ItemsSelection/AddonsItems.tsx
```diff
@@ -1,17 +1,92 @@
...
       color={color("grey", 40)}
     />
+  ) : isShowAddonItems ? (
+    <>
+      {transactionList && transactionList.length > 0
+        ? transactionList.map(
+            (transactionsByCategory: TransactionsGroup, index: number) => (
+              <TransactionsGroup
+                key={index}
+                maxTransactionsToDisplay={!isShowAddonItems ? 0 : BIG_NUMBER}
```
sebastianquek, line 81: This can now always be set to `BIG_NUMBER` since `isShowAddonItems` will be true

src/components/CustomerQuota/ItemsSelection/AddonsItems.tsx
```diff
@@ -1,17 +1,92 @@
...
+                {...transactionsByCategory}
+                header={undefined}
```
sebastianquek, line 83: Can omit this prop since it's undefined

src/components/CustomerQuota/ItemsSelection/AddonsItems.tsx
```diff
@@ -1,17 +1,92 @@
...
+              />
+            )
+          )
+        : undefined}
```
sebastianquek, line 87: Small nit, noticed you're using `undefined` to notate no rendering of items, shall we use `null` instead so we keep it consistent across the app?

src/components/CustomerQuota/ItemsSelection/AddonsItems.tsx
andrehadianto: as discussed on slack, will be kept to replace `header` from `{...transactionsByCategory}`

sebastianquek (approved):
LGTM!! thanks for making all the updates!

PR #307, opened 2021-02-17
fix: cart quota not updating

yongjiajie:
[Notion link](https://www.notion.so/When-selecting-2-NRICs-in-Engineering-Good-campaign-the-checkbox-does-not-turn-into-a-stepper-826f45042f4e4b3b815bb8341f22e3ea) 

This PR adds... 

- Cart quota now updates whenever the quota response changes

### DIFF
--- a/src/hooks/useCart/useCart.tsx
+++ b/src/hooks/useCart/useCart.tsx
@@ -133,7 +133,8 @@ export const useCart = (
   useEffect(() => {
     if (
       cartQuota &&
-      (prevCartQuota != cartQuota ||
+      (!prevCartQuota ||
+        prevCartQuota != cartQuota ||
         prevIds !== ids ||
         prevProducts !== products)
     ) {
--- a/src/hooks/useCart/useCart.tsx
+++ b/src/hooks/useCart/useCart.tsx
@@ -125,10 +125,11 @@ export const useCart = (
 
   /**
    * Update the cart when:
-   *  1. First quota is retrieved on initialisation (i.e. no previous cart quota)
-   *  2. When quota response changes
-   *  3. When quota is updated (i.e. products or ids change)
-   * Items 2 and 3 can occur at the same time.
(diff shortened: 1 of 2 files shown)


src/hooks/useCart/useCart.tsx
```diff
@@ -133,7 +133,9 @@ export const useCart = (
   useEffect(() => {
     if (
       cartQuota &&
-      (!prevCartQuota || prevIds !== ids || prevProducts !== products)
+      (prevCartQuota != cartQuota ||
```
ilhaoni, line 136: hmm curious, do you think we should keep the original `!prevCartQuota` and add this new line of logic?

Would there be issues if we keep only `prevCartQuota != cartQuota` and 1st quota is not retrievable upon initialisation 🤔

ilhaoni (commented):
Thanks for making this fix!! :D 

Just have one question haha

src/hooks/useCart/useCart.tsx
yongjiajie: I don't think there will be a case where `cartQuota` will be `undefined` for that condition to be `true`, since if it fails to fetch the quota properly, quota errors will be thrown. However, I do think that having the condition of `!prevCartQuota` reduces all ambiguity, and is a better safeguard. Good catch!

The conditions for updating the cart should be:

1. An incoming cart quota exists, AND
2. There is no existing cart quota, OR
3. The incoming cart quota is different from the existing cart quota, OR
4. The customer IDs have changed, OR
5. The products in the cart have changed

Adding the condition back should make the above clearer.

ilhaoni: sounds good~ :D 

small nit, shall we update the conditions in the comment above also? 😆

ilhaoni (approved):
LGTM! :D 

Thanks for helping to fix this~~~

PR #310, opened 2021-02-22
chore(past-tx): move the logging of IDs to a breadcrumb

sebastianquek:
This ensures that the errors are logged in a single category on Sentry. Now they are scattered across many different categories.

### DIFF
--- a/src/hooks/usePastTransaction/usePastTransaction.tsx
+++ b/src/hooks/usePastTransaction/usePastTransaction.tsx
@@ -31,6 +31,12 @@ export const usePastTransaction = (
 
   useEffect(() => {
     const fetchPastTransactions = async (): Promise<void> => {
+      Sentry.addBreadcrumb({
+        category: "fetchPastTransactions",
+        message: JSON.stringify({
+          ids: ids.map((id) => id.slice(-4).padStart(id.length, "*")),
+        }),
+      });
       try {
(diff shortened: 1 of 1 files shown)


src/hooks/usePastTransaction/usePastTransaction.tsx
```diff
@@ -31,6 +31,12 @@ export const usePastTransaction = (
 
   useEffect(() => {
     const fetchPastTransactions = async (): Promise<void> => {
+      Sentry.addBreadcrumb({
```
szenius, line 34: Not very familiar with Sentry so have to ask - what is the difference between `addBreadcrumb` and `captureException`?

sebastianquek: `captureException` is used as the final submission of an error, while `addBreadcrumb` is used to add specific info so that we can trace back the steps the user took before the error was raised.

For this particular case, we were using `captureException` with a string (typically it's an Error), and Sentry treats each error type as its own issue so that we can see as a whole how many errors of that particular type was raised. This PR leaves it as a string, but removes the dynamic portion (i.e. masked NRIC) so that the errors are combined under the same issue.

szenius: Thank you for explaining!!

szenius (approved):
Thank you for doing this fix!

PR #320, opened 2021-03-05
fix: add checks for empty array in hasNoQuota

yongjiajie:
This PR adds... 

- An additional check in `hasNoQuota` to check for empty whether `remainingQuota` is empty before running `every` on it, since it returns true if the array is empty.

### Notes 

Note that we want `hasNoQuota` to return true when `remainingQuota` is either:
- empty, or
- all of the items have zero quantities

### DIFF
--- a/src/hooks/useQuota/useQuota.tsx
+++ b/src/hooks/useQuota/useQuota.tsx
@@ -74,10 +74,7 @@ const filterQuotaWithAvailableProducts = (
  * @returns true if there is no quota, otherwise false.
  */
 const hasNoQuota = (quota: Quota): boolean => {
-  return (
-    quota.remainingQuota.length <= 0 ||
-    quota.remainingQuota.every((item) => item.quantity === 0)
-  );
+  return quota.remainingQuota.every((item) => item.quantity === 0);
 };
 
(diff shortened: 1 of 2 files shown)


src/hooks/useQuota/useQuota.tsx
```diff
@@ -63,8 +63,21 @@ const filterQuotaWithAvailableProducts = (
...
+ *
+ * @param quota a Quota object
+ * @returns true if there is no quota, otherwise false.
+ */
 const hasNoQuota = (quota: Quota): boolean => {
-  return quota.remainingQuota.every((item) => item.quantity === 0);
+  return (
+    quota.remainingQuota.length <= 0 ||
+    quota.remainingQuota.every((item) => item.quantity === 0)
+  );
```
szenius, line 80: Hmm like what seb mentioned, this is same as the original condition, so actually no need to add line 78 le~~ The main point for this PR probably is for the tests!

szenius: I think the docs are good though, so it lets us know it is intended

szenius (approved):
Nice thank you for improving the docs and testing 🙌

PR #324, opened 2021-03-16
feat: create unique accessibility labels for each identifier

ilhaoni:
This PR uses label in accessibilityLabel for IdentifierTextInput and IdentifierPhoneNumberInput. This allows us to differentiate each identifierInput, as each of them will have a unique accessibilityLabel. This would help the e2e tests and reduce reliance on indexes and order of the identifiers.

Current behaviour:
- all text input type that is `PHONE_NUMBER` will have an accessibility label of `item-field-phone-number`
- all other text input type will have an accessibility label of `item-field-text`

New behaviour:
- for text input type with `PHONE_NUMBER`, if label is present, we will now format as `<label>-phone-number`. E.g. `Main Contact` will result in `main-contact-phone-number`
- for all other text input, we will now format as `<label>-text`. E.g. `Designation` will result in `designation-text`
- if label is not present, the above default values will be used isntead.

This PR adds... 

- [ ] I've kept this PR as small as possible (~600 lines) by splitting it into PRs with manageable chunks of code

### DIFF
--- a/src/components/CustomerQuota/ItemsSelection/IdentifierLayout/IdentifierPhoneNumberInput.tsx
+++ b/src/components/CustomerQuota/ItemsSelection/IdentifierLayout/IdentifierPhoneNumberInput.tsx
@@ -61,7 +61,15 @@ export const IdentifierPhoneNumberInput: FunctionComponent<{
         mobileNumberValue={phoneNumber}
         onChangeCountryCode={setCountryCodeValue}
         onChangeMobileNumber={setPhoneNumber}
-        accessibilityLabel="item-field-phone-number"
+        accessibilityLabel={
+          label
+            ? label
+                .trim()
+                .replace(/\s/g, "-")
+                .toLowerCase()
+                .concat("-phone-number")
+            : "item-field-phone-number"
+        }
       />
     </View>
   );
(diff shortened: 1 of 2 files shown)


yongjiajie (commented):
Looks good! Would it make sense to extract this logic into a utility function we can use in the future? Or do we forsee not using this a lot?

andrehadianto (changes requested):
should we add `testId` as well? Since the document mentioned that `testId` is used for iOS while `accessibilityLabel` is used for Android

ilhaoni:
Hmm @yongjiajie currently we know for sure that identifiers have labels, so I was thinking we can push this first, and then we examine how to create a utility function 🤔  what do you think about this haha. Feels very trial and error at the moment 😂 

So far, only identifiers seem to have the same accessibility ids [here](https://www.notion.so/Testing-Add-essential-accessibility-IDs-3c0fd1ebab2041a5b0312664b7299000). I guess the next ones to tackle will be 
- logout related 
- distribution name
- stepper related
- checkbox related

@andrehadianto ahh, it does pass down to `testId`! 
- `IdentifierPhoneNumberInput` uses `PhoneNumberInput`, which generates all the `accessibilityLabel` for Android and `testID` for iOS haha
-  `IdentifierTextInput` uses `InputWithLabel`, which generates all the `accessibilityLabel` for Android and `testID` for iOS haha

yongjiajie (approved):
Sounds good, I thought it was something consistent! Thanks for sharing the context 🎉

andrehadianto (approved):

PR #359, opened 2021-05-07
hotfix: wrong condition logic

andrehadianto:

This PR adds... 
- replace check barcode logic to check scannerIdType label

### DIFF
--- a/src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
+++ b/src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
@@ -203,11 +203,11 @@ const CollectCustomerDetailsScreen: FunctionComponent<NavigationFocusInjectedPro
 
   const onBarCodeScanned: BarCodeScannedCallback = (event) => {
     if (isFocused && isScanningEnabled && event.data) {
-      if (event.type === BarCodeScanner.Constants.BarCodeType.code39) {
-        onCheck(event.data);
-      } else {
+      if (selectedIdType.label === "Passport") {
         const { passportId } = JSON.parse(event.data);
         onCheck(passportId);
+      } else {
+        onCheck(event.data);
       }
     }
   };
--- a/src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
+++ b/src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
@@ -47,7 +47,6 @@ import {
   defaultSelectedIdType,
 } from "../../context/identification";
 import { useTranslate } from "../../hooks/useTranslate/useTranslate";
-import { NetworkError } from "../../services/helpers";
 
 const styles = StyleSheet.create({
   content: {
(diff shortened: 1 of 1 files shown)


src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
```diff
@@ -173,6 +173,25 @@ const CollectCustomerDetailsScreen: FunctionComponent<NavigationFocusInjectedPro
...
+     */
+    if (selectedIdType.label === "Passport") {
+      let passportId;
+      try {
+        ({ passportId } = JSON.parse(input));
+      } catch (e) {
+        /**
+         * We set this to empty so an appropriate error dialog will be
+         * shown, instead of an ErrorBoundary or empty dialog.
+         */
```
dalsontws, line 188: wah interesting, so after leaving this empty, how is the error handled?

ilhaoni: is it caught by error boundary ah? do we need to throw out the error again after catching it?

ilhaoni (commented):
reviewed! just have a similar question with Dalson, on how the error scenarios are handled

src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
yongjiajie: This error is thrown because it's not a valid JSON object, so there's no `passportId` to parse. We leave it empty so that the error is handled below, similar to the case when there's no input at all.

dalsontws: ok understood! in that case, will it be better for us to catch the syntax error (i.e. `catch (e instanceof SyntaxError)` instead of catching all errors, so that it's clearer that this block is use specifically for json parsing errors?

in the event there are any unforeseen errors, it will go to the error boundary instead

src/components/CustomerDetails/CollectCustomerDetailsScreen.tsx
```diff
@@ -172,7 +172,35 @@ const CollectCustomerDetailsScreen: FunctionComponent<NavigationFocusInjectedPro
...
+        /**
+         * If we encounter any JSON errors, we set `passportId` to empty
+         * in order to allow logic below to handle it as an empty/invalid ID.
+         *
+         * However, any other errors will trigger the ErrorBoundary.
+         */
+        if (e instanceof SyntaxError) {
+          passportId = "";
+        } else {
+          setState(() => {
```
andrehadianto, line 195: can't seem to find the answer on google 😢 .
What `state` is this update function changing? Since you only retrieve `useState()[1]` in line 175

yongjiajie: Since we want the error to be thrown to the error boundary instead, we'll have to update React this way since React error boundaries don't actually catch errors in async code. You can see [this](https://reactjs.org/docs/error-boundaries.html)!

ilhaoni (approved):
Tested~ works!!

Thanks for doing the hotfix! \^^/

dalsontws (approved):
LGTM!  🚀

PR #371, opened 2021-05-28
feat: PAYMENT_RECEIPT to have optional value.

andrehadianto:

This PR adds... 
- remove mandatory `value` for `PAYMENT_RECEIPT`
- update `PolicyIdentifier` type

I think 
``` typescript
    if (textInputType !== "PAYMENT_RECEIPT" && !value) {
      throw new Error(ERROR_MESSAGE.MISSING_IDENTIFIER_INPUT);
    }
```
can be removed instead. Empty `value` will still be checked in the `regex` matching as well as BE validation

- [x] I've added/updated [e2e tests](https://www.notion.so/e2e-Documentation-a096d3e0bf75485e85ad692af8371ef3) for at least the happy flow

### DIFF
--- a/src/utils/validateIdentifierInputs.ts
+++ b/src/utils/validateIdentifierInputs.ts
@@ -58,6 +58,5 @@ export const validateIdentifierInputs = (
   ) {
     throw new Error(ERROR_MESSAGE.DUPLICATE_IDENTIFIER_INPUT);
   }
-  console.log("asdfasdf");
   return true;
 };
--- a/src/utils/validateIdentifierInputs.ts
+++ b/src/utils/validateIdentifierInputs.ts
@@ -19,24 +19,20 @@ export const validateIdentifierInputs = (
   identifierInputs: IdentifierInput[]
 ): boolean => {
   for (const { value, validationRegex, textInputType } of identifierInputs) {
-    if (
-      textInputType === "PAYMENT_RECEIPT" &&
-      (!value ||
-        !isMatchRegex(
-          value,
-          validationRegex
-            ? validationRegex
-            : defaultPaymentReceiptValidationRegex
-        ))
-    ) {
-      throw new Error(ERROR_MESSAGE.INVALID_PAYMENT_RECEIPT_NUMBER);
-    }
-    if (!value) {
+    if (textInputType !== "PAYMENT_RECEIPT" && !value) {
       throw new Error(ERROR_MESSAGE.MISSING_IDENTIFIER_INPUT);
     }
-
     if (textInputType === "NUMBER" && isNaN(Number(value))) {
       throw new Error(ERROR_MESSAGE.INVALID_IDENTIFIER_INPUT);
+    }
+    if (
+      textInputType === "PAYMENT_RECEIPT" &&
+      !isMatchRegex(
+        value,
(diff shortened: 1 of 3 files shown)


src/utils/validateIdentifierInputs.ts
```diff
@@ -62,6 +58,6 @@ export const validateIdentifierInputs = (
   ) {
     throw new Error(ERROR_MESSAGE.DUPLICATE_IDENTIFIER_INPUT);
   }
-
+  console.log("asdfasdf");
```
dalsontws, line 61: need to remove the console.log here 😆

src/utils/validateIdentifierInputs.ts
```diff
@@ -19,25 +19,21 @@ export const validateIdentifierInputs = (
...
-      textInputType === "PAYMENT_RECEIPT" &&
-      (!value ||
-        !isMatchRegex(
-          value,
-          validationRegex
-            ? validationRegex
-            : defaultPaymentReceiptValidationRegex
-        ))
-    ) {
-      throw new Error(ERROR_MESSAGE.INVALID_PAYMENT_RECEIPT_NUMBER);
```
dalsontws, line 32: hmm actually this part is still required (can just remove the `!value` check), as invalid payment receipt will throw a specific error message, see `ERROR_MESSAGE.INVALID_PAYMENT_RECEIPT_NUMBER`

dalsontws (changes requested):

andrehadianto (commented):
press start review by mistake :|

src/utils/validateIdentifierInputs.ts
dalsontws: ah okay, understand! yup i think this will work, should we include a test case to check for empty payment receipt fields?

src/utils/validateIdentifierInputs.ts
```diff
@@ -19,25 +19,21 @@ export const validateIdentifierInputs = (
...
-    }
-    if (!value) {
+    if (textInputType !== "PAYMENT_RECEIPT" && !value) {
```
dalsontws, line 22: realised that altho the backend allows for optional fields for different text inputs types, this current logic only allows optional payment receipts. shall we address this in another story?

andrehadianto: sounds good! But the PR of that story would be the same as the PR676?

ilhaoni: or shall we address this inside PR676 haha

ilhaoni: actually another question, is there a reason why we're doing an additional check here of  whether `textInputType !== "PAYMENT_RECEIPT"` haha

ilhaoni (commented):
Left some comments!

src/utils/validateIdentifierInputs.ts
andrehadianto: as discussed,, need the check cause we only allow `PAYMENT_RECEIPT` to be empty currently!

ilhaoni: ok!

ilhaoni (approved):
LGTM! :D

dalsontws (approved):
Thanks for working on this, LGTM!🚀

PR #377, opened 2021-06-16
fix: validate regex input for leading and trailing whitespaces

findtravishere:
[Notion link](https://www.notion.so/Trim-extra-leading-or-trailing-spaces-from-user-input-738d72c9467740d9b5e77b3b4828cf28) 

This PR adds the fix for trailing and leading whitespace for non NRIC signup values. 

### DIFF
--- a/src/hooks/useCart/useCart.test.tsx
+++ b/src/hooks/useCart/useCart.test.tsx
@@ -586,9 +586,9 @@ describe("useCart", () => {
         ]);
         result.current.updateCart("chocolate", 5, [
           {
-            label: "last",
-            value: "        last      ",
-            textInputType: "STRING",
+            label: "code",
+            value: "        +6581898380      ",
+            textInputType: "PHONE_NUMBER",
           },
         ]);
       });
@@ -621,9 +621,9 @@ describe("useCart", () => {
           descriptionAlert: undefined,
           identifierInputs: [
             {
-              label: "last",
-              textInputType: "STRING",
-              value: "last",
+              label: "code",
+              textInputType: "PHONE_NUMBER",
+              value: "+6581898380",
             },
           ],
           lastTransactionTime: transactionTime,
@@ -680,6 +680,67 @@ describe("useCart", () => {
           {
             label: "second",
             value: "second",
+            textInputType: "PAYMENT_RECEIPT",
+          },
+        ]);
+        result.current.checkoutCart();
+      });
+
+      expect(result.current.cartState).toBe("PENDING_CONFIRMATION");
+      expect(result.current.cart).toStrictEqual([
+        {
+          category: "toilet-paper",
+          descriptionAlert: undefined,
+          identifierInputs: [
+            {
+              label: "second",
+              textInputType: "PAYMENT_RECEIPT",
+              value: "second",
+            },
+          ],
+          lastTransactionTime: transactionTime,
+          maxQuantity: 2,
+          quantity: 1,
+        },
+        {
+          category: "chocolate",
+          descriptionAlert: undefined,
+          identifierInputs: [
+            {
+              label: "first",
+              scanButtonType: "BARCODE",
+              textInputType: "STRING",
+              value: "",
+            },
+            {
+              label: "last",
+              scanButtonType: "BARCODE",
+              textInputType: "STRING",
+              value: "",
+            },
+          ],
+          lastTransactionTime: transactionTime,
+          maxQuantity: 15,
+          quantity: 0,
+        },
+      ]);
+    });
+
+    it("should show PENDING_CONFIRMATION cart state when a PAYMENT_RECEIPT identifier input exists with whitespaced values", async () => {
+      expect.assertions(2);
+      const ids = ["ID1"];
+      const { result } = renderHook(
+        () => useCart(ids, key, endpoint, mockQuotaResSingleId.remainingQuota),
+        {
+          wrapper,
+        }
+      );
+
+      await waitFor(() => {
+        result.current.updateCart("toilet-paper", 1, [
+          {
+            label: "second",
+            value: "         second           ",
             textInputType: "PAYMENT_RECEIPT",
           },
         ]);
--- a/src/hooks/useCart/useCart.test.tsx
+++ b/src/hooks/useCart/useCart.test.tsx
@@ -566,6 +566,76 @@ describe("useCart", () => {
       );
     });
 
+    it("should set the correct checkoutResult when checkoutCart is called with whitespaced values", async () => {
+      expect.assertions(4);
+      const ids = ["ID1"];
+      const { result } = renderHook(
+        () => useCart(ids, key, endpoint, mockQuotaResSingleId.remainingQuota),
+        {
+          wrapper,
+        }
+      );
+
+      await waitFor(() => {
+        result.current.updateCart("toilet-paper", 2, [
+          {
+            label: "first",
+            value: "      first          ",
+            textInputType: "STRING",
+          },
+        ]);
+        result.current.updateCart("chocolate", 5, [
+          {
+            label: "last",
+            value: "        last      ",
+            textInputType: "STRING",
+          },
+        ]);
+      });
+
+      mockPostTransaction.mockReturnValueOnce(mockPostTransactionResult);
+
+      await waitFor(() => {
+        result.current.checkoutCart();
+        expect(result.current.cartState).toBe("CHECKING_OUT");
+      });
+
+      expect(result.current.cartState).toBe("PURCHASED");
+      expect(result.current.cart).toStrictEqual([
+        {
+          category: "toilet-paper",
+          descriptionAlert: undefined,
+          identifierInputs: [
+            {
+              label: "first",
+              textInputType: "STRING",
+              value: "first",
+            },
+          ],
+          lastTransactionTime: transactionTime,
+          maxQuantity: 2,
+          quantity: 2,
+        },
+        {
+          category: "chocolate",
+          descriptionAlert: undefined,
+          identifierInputs: [
+            {
+              label: "last",
+              textInputType: "STRING",
+              value: "last",
+            },
+          ],
+          lastTransactionTime: transactionTime,
+          maxQuantity: 15,
+          quantity: 5,
+        },
+      ]);
+      expect(result.current.checkoutResult).toStrictEqual(
+        mockPostTransactionResult
+      );
+    });
+
     it("should show unsuccessful cart state when token mismatch error is received (return-pod)", async () => {
       expect.assertions(2);
       const ids = ["ID1"];
--- a/src/hooks/useCart/useCart.test.tsx
+++ b/src/hooks/useCart/useCart.test.tsx
@@ -587,7 +587,7 @@ describe("useCart", () => {
         result.current.updateCart("chocolate", 5, [
           {
             label: "code",
-            value: "        +6581898380      ",
+            value: "        +6588888888      ",
             textInputType: "PHONE_NUMBER",
           },
         ]);
@@ -623,7 +623,7 @@ describe("useCart", () => {
             {
               label: "code",
               textInputType: "PHONE_NUMBER",
-              value: "+6581898380",
+              value: "+6588888888",
             },
           ],
           lastTransactionTime: transactionTime,
@@ -744,6 +744,20 @@ describe("useCart", () => {
             textInputType: "PAYMENT_RECEIPT",
           },
         ]);
+        result.current.updateCart("chocolate", 4, [
+          {
+            value: "        first             ",
+            label: "first",
+            textInputType: "STRING",
+            scanButtonType: "BARCODE",
+          },
+          {
+            value: "       last       ",
+            label: "last",
+            textInputType: "STRING",
+            scanButtonType: "BARCODE",
+          },
+        ]);
         result.current.checkoutCart();
       });
 
@@ -771,18 +785,18 @@ describe("useCart", () => {
               label: "first",
               scanButtonType: "BARCODE",
               textInputType: "STRING",
-              value: "",
+              value: "first",
             },
             {
               label: "last",
               scanButtonType: "BARCODE",
               textInputType: "STRING",
-              value: "",
+              value: "last",
             },
           ],
           lastTransactionTime: transactionTime,
           maxQuantity: 15,
-          quantity: 0,
+          quantity: 4,
         },
       ]);
     });
--- a/src/hooks/useCart/useCart.tsx
+++ b/src/hooks/useCart/useCart.tsx
@@ -1,7 +1,8 @@
 import { useState, useCallback, useContext, useEffect } from "react";
 import { postTransaction } from "../../services/quota";
 import { PostTransactionResult, ItemQuota, IdentifierInput } from "../../types";
-import { validateIdentifierInputs } from "../../utils/validateAndCleanIdentifierInputs";
+import { validateIdentifierInputs } from "../../utils/validateIdentifierInputs";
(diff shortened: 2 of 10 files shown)


src/utils/validateInputWithRegex.ts
```diff
@@ -6,8 +6,8 @@ export const validateAndCleanRegexInput = (
   inputId: string,
   idRegex: string
 ): string => {
-  // set ID to all uppercase to remove case sensitivity
-  const id = inputId.toUpperCase();
+  // set ID to all uppercase to remove case sensitivity and removes trailing and leading white spaces
+  const id = inputId.trim().slice(0).toUpperCase();
```
dalsontws, line 10: hmm whats the reason for including slice here? 
this might not be applicable, do take a look at the documentation: https://www.w3schools.com/jsref/jsref_slice_string.asp

dalsontws: also! do add test cases for this cases, to ensure the behaviour is intended

findtravishere: i added the slice since just to ensure we're returning a new copy of it, should I just remove it?

findtravishere: added test cases!

dalsontws: okay i think it's fine to include the slice. also do make sure to perform testing on the distributions where IDs are in regex form (i.e. FFTH, Haze) to ensure the behaviour is as intended!

src/utils/validateIdentifierInputs.ts
```diff
@@ -43,7 +43,7 @@ export const validateIdentifierInputs = (
     ) {
       throw new Error(ERROR_MESSAGE.INVALID_PHONE_AND_COUNTRY_CODE);
     }
-    if (!isMatchRegex(value, validationRegex)) {
+    if (!isMatchRegex(value.trim(), validationRegex)) {
       throw new Error(ERROR_MESSAGE.INVALID_IDENTIFIER_INPUT);
     }
     if (textInputType === "PHONE_NUMBER" && !fullPhoneNumberValidator(value)) {
```
dalsontws, line 49: any reason why we are not calling `trim()` for the phone number?

src/utils/validateIdentifierInputs.ts
```diff
@@ -28,7 +28,7 @@ export const validateIdentifierInputs = (
     if (
       textInputType === "PAYMENT_RECEIPT" &&
       !isMatchRegex(
-        value,
+        value.trim(),
```
dalsontws, line 31: noticed we are calling `trim()` multiple times across this function, do you think it's better to trim the value at the start?

findtravishere: updated to trim all values from the start!

src/utils/validateIdentifierInputs.ts
```diff
@@ -19,19 +19,20 @@ export const validateIdentifierInputs = (
   identifierInputs: IdentifierInput[]
 ): boolean => {
   for (const { value, validationRegex, textInputType } of identifierInputs) {
-    if (textInputType === "SINGLE_CHOICE" && !value) {
+    const trimmedValue = value.trim();
```
dalsontws, line 22: notice the first 2 `if` conditions are checking if `value` exist. since there's a possibility that `value` is undefined, do you think we need to passed the trimmed value in here?

findtravishere: didn't catch this while changing all the values, thanks, updated!

src/utils/validateIdentifierInputs.ts
findtravishere: updated!

src/utils/validateIdentifierInputs.ts
dalsontws: no problem! do you think it will be better to move the value.trim() to after these checks as well? since it's not used for the 3 conditions!

findtravishere: moved it! i thought about doing that but I left it there because I thought it might be used somewhere in the checks in the future

src/utils/validateInputWithRegex.ts
sebastianquek: hello, I'd suggest to remove the slice, I came across this, was wondering why. Since `inputId` is a primitive string, don't need to worry about copying

image

sebastianquek:
There's a subtle issue, but requires quite a bit of change. Although you've fixed `validateIdentifierInputs`, look at its call hierarchy (right click the fn, Show Call Hierarchy). Notice that it's called in `checkoutCart`. The problem is that although the identifier inputs are validated, when we send it over to the backend, it's not properly cleaned and trimmed.

A good example is to look at the `validateAndCleanId` function. It ensures the id that is provided by the user is properly cleaned, then subsequent functions that act upon this id already uses the cleaned version.

src/hooks/useCart/useCart.tsx
```diff
@@ -233,6 +233,9 @@ export const useCart = (
       const transactions = Object.values(cart)
         .filter(({ quantity }) => quantity)
         .map(({ category, quantity, identifierInputs }) => {
+          for (const identifierInput of identifierInputs) {
```
dalsontws, line 236: notice this additional loop to trim the identifier inputs would work, but we will be looping through them twice; once here, and another time at line 250. do you think we can validate and clean the identifier inputs together here?

src/hooks/useCart/useCart.tsx
```diff
@@ -235,11 +235,22 @@ export const useCart = (
...
+      const allIdentifierInputs: IdentifierInput[] = [];
+      let transactions: any = [];
+      try {
+        transactions = Object.values(cart)
+          .filter(({ quantity }) => quantity)
+          .map(({ category, quantity, identifierInputs }) => {
+            allIdentifierInputs.push(...identifierInputs);
+            validateIdentifierInputs(allIdentifierInputs);
+            return { category, quantity, identifierInputs };
+          });
```
dalsontws, line 247: Notice you are still returning the uncleaned identifierInputs here (i.e. `return { category, quantity, identifierInputs };`). In addition, this logic will result to the validation of identiferInputs of the previous transactions repeatedly. 

how about you try smth like this:

```suggestion
      try {
        transactions = Object.values(cart)
          .filter(({ quantity }) => quantity)
          .map(({ category, quantity, identifierInputs }) => {
            const cleanedIdentifierInputs = validateIdentifierInputs(identifierInputs);
            return { category, quantity, identifierInputs: cleanedIdentifierInputs };
          });
```
do add a test in `useCart.test.tsx` to cover this change, and make sure whitespaced inputs are not passed in

src/hooks/useCart/useCart.tsx
```diff
@@ -297,26 +308,27 @@ export const useCart = (
...
-        });
-
-      if (transactions.length === 0) {
+      let transactions: any = [];
+      try {
+        transactions = Object.values(cart)
+          .filter(({ quantity }) => quantity)
+          .map(({ category, quantity, identifierInputs }) => {
+            allIdentifierInputs.push(...identifierInputs);
+            validateIdentifierInputs(allIdentifierInputs);
```
dalsontws, line 317: do make the same change here as well!

src/hooks/useCart/useCart.tsx
```diff
@@ -242,8 +242,10 @@ export const useCart = (
           .filter(({ quantity }) => quantity)
           .map(({ category, quantity, identifierInputs }) => {
             allIdentifierInputs.push(...identifierInputs);
-            validateIdentifierInputs(allIdentifierInputs);
-            return { category, quantity, identifierInputs };
+            const cleanedIdentifierInputs = validateIdentifierInputs(
+              allIdentifierInputs
```
dalsontws, line 246: do we still need line 244 to push all the `identifierInputs` in a single array?

src/hooks/useCart/useCart.tsx
```diff
@@ -296,28 +315,36 @@ export const useCart = (
...
         setCartState("DEFAULT");
-        setCartError(new Error(ERROR_MESSAGE.MISSING_SELECTION));
+        setCartError(error);
         return;
       }
 
-      try {
-        validateIdentifierInputs(allIdentifierInputs);
-      } catch (error) {
+      if (transactions.length === 0) {
```
dalsontws, line 341: do move this conditional check within the try catch above (line 336), since the `validateIdentifierInputs` is now inside that.

src/hooks/useCart/useCart.tsx
```diff
@@ -235,11 +236,29 @@ export const useCart = (
...
+            allCleanedIdentifierInputs.push(...cleanedIdentifierInputs);
+            return {
+              category,
+              quantity,
+              identifierInputs: cleanedIdentifierInputs,
+            };
+          });
+        validateIdentifierInputs(allCleanedIdentifierInputs);
+      } catch (error) {
+        setCartState("DEFAULT");
```
dalsontws, line 257: do add the check for empty transactions here too, and subsequently throw the error message `new Error(ERROR_MESSAGE.MISSING_SELECTION))`

src/hooks/useCart/useCart.test.tsx
```diff
@@ -566,6 +566,76 @@ describe("useCart", () => {
...
+            label: "first",
+            value: "      first          ",
+            textInputType: "STRING",
+          },
+        ]);
+        result.current.updateCart("chocolate", 5, [
+          {
+            label: "last",
+            value: "        last      ",
+            textInputType: "STRING",
```
dalsontws, line 591: nit: do all other textInput types here instead of just the STRING type!

src/hooks/useCart/useCart.tsx
findtravishere: done!

src/hooks/useCart/useCart.tsx
findtravishere: done!

src/hooks/useCart/useCart.test.tsx
```diff
@@ -656,6 +726,67 @@ describe("useCart", () => {
...
+      const ids = ["ID1"];
+      const { result } = renderHook(
+        () => useCart(ids, key, endpoint, mockQuotaResSingleId.remainingQuota),
+        {
+          wrapper,
+        }
+      );
+
+      await waitFor(() => {
+        result.current.updateCart("toilet-paper", 1, [
```
dalsontws, line 740: do you want to add the "chocolate" categories here with whitespaced values? so that it wont be empty for results below

src/hooks/useCart/useCart.tsx
```diff
@@ -235,11 +236,22 @@ export const useCart = (
...
+            identifierInputs
+          );
+          allCleanedIdentifierInputs.push(...cleanedIdentifierInputs);
+          return {
+            category,
+            quantity,
+            identifierInputs: cleanedIdentifierInputs,
+          };
         });
+
```
dalsontws, line 254: nit: maybe can remove the additional empty lines!

src/hooks/useCart/useCart.tsx
```diff
@@ -235,11 +236,22 @@ export const useCart = (
    */
   const _completeCheckout = useCallback(() => {
     const complete = async (): Promise<void> => {
-      const transactions = Object.values(cart)
+      const allCleanedIdentifierInputs: IdentifierInput[] = [];
+      let transactions: any = [];
```
dalsontws, line 240: do we still need to declare the transactions here? or can we just use `const transactions` in the next line?

src/utils/validateIdentifierInputs.ts
```diff
@@ -61,5 +67,6 @@ export const validateIdentifierInputs = (
   ) {
     throw new Error(ERROR_MESSAGE.DUPLICATE_IDENTIFIER_INPUT);
   }
+
```
dalsontws, line 70: nit: same here for the empty line

src/hooks/useCart/useCart.tsx
findtravishere: fixed!

src/hooks/useCart/useCart.test.tsx
findtravishere: added!

src/utils/validateIdentifierInputs.ts
findtravishere: done!

src/hooks/useCart/useCart.tsx
findtravishere: done!

src/hooks/useCart/useCart.tsx
```diff
@@ -235,10 +236,19 @@ export const useCart = (
...
     const complete = async (): Promise<void> => {
+      const allCleanedIdentifierInputs: IdentifierInput[] = [];
       const transactions = Object.values(cart)
         .filter(({ quantity }) => quantity)
         .map(({ category, quantity, identifierInputs }) => {
-          return { category, quantity, identifierInputs };
+          const cleanedIdentifierInputs = cleanIdentifierInputs(
+            identifierInputs
+          );
+          allCleanedIdentifierInputs.push(...cleanedIdentifierInputs);
```
dalsontws, line 246: looks like we no longer need `allCleanedIdentifierInputs` here since it's not validated in the `completeCheckout` fn!

src/hooks/useCart/useCart.tsx
```diff
@@ -235,10 +236,19 @@ export const useCart = (
    */
   const _completeCheckout = useCallback(() => {
     const complete = async (): Promise<void> => {
+      const allCleanedIdentifierInputs: IdentifierInput[] = [];
```
dalsontws, line 239: can remove this too!

src/hooks/useCart/useCart.tsx
```diff
@@ -295,13 +305,19 @@ export const useCart = (
   const checkoutCart: CartHook["checkoutCart"] = useCallback(() => {
     const checkout = async (): Promise<void> => {
       setCartState("CHECKING_OUT");
-
-      const allIdentifierInputs: IdentifierInput[] = [];
-      const transactions = Object.values(cart)
+      const allCleanedIdentifierInputs: IdentifierInput[] = [];
+      const transactions: any = Object.values(cart)
```
dalsontws, line 309: do we need to define transactions as `any`?

src/utils/validateIdentifierInputs.ts
```diff
@@ -28,10 +28,13 @@ export const validateIdentifierInputs = (
     if (textInputType === "NUMBER" && isNaN(Number(value))) {
       throw new Error(ERROR_MESSAGE.INVALID_IDENTIFIER_INPUT);
     }
+
+    const trimmedValue = value.trim();
```
dalsontws, line 32: we no longer need to trim here since it's already trimmed in `cleanIdentifierInputs`!! can remove this

src/hooks/useCart/useCart.tsx
findtravishere: removed!

src/hooks/useCart/useCart.tsx
findtravishere: removed!

src/utils/validateIdentifierInputs.test.ts
```diff
@@ -1,7 +1,7 @@
 import { validateIdentifierInputs } from "./validateIdentifierInputs";
 
 describe("validateIdentifierInputs", () => {
-  it("should return true if all identifiers are valid", () => {
+  it("should return validated input if all identifiers are valid", () => {
```
dalsontws, line 4: do revert this to true 
```suggestion
  it("should return true if all identifiers are valid", () => {
```

dalsontws (approved):
Thanks for making all the changes! LGTM 🚀

PR #398, opened 2021-07-16
feat: use securestore for tokens

jedidiahkkh:
[Plaintext token storage in unencrypted database files (PT-M004)](https://www.notion.so/Plaintext-token-storage-in-unencrypted-database-files-PT-M004-3497074e9e474f32bcd44c9c7add63e1) 

Changes storage of secrets (operator token, session token, endpoints) to SecureStore which is encrypted instead of Asyncstorage that is unencrypted.

This PR is for the changes to AuthStore components
* AuthStoreProvider

Changes for configStore will be in a separate PR

- [x] (>.< sorry this is slightly more than 600 but it's as small as possible) I've kept this PR as small as possible (~600 lines) by splitting it into PRs with manageable chunks of code
- [ ] (no new components created) I've added [accessibility IDs](https://www.notion.so/dc5f4ce910f7431c84344ac79344e9f5?v=d487366816834dd4ab8dc12e0b5928f3) to my components
- [ ] (not needed as no user flow changes) I've added/updated [e2e tests](https://www.notion.so/e2e-Documentation-a096d3e0bf75485e85ad692af8371ef3) for at least the happy flow
- [ ] (not needed as no UI changes) I've added [Chinese translations for i18n setup](https://www.notion.so/Translations-Dev-Guide-8c1da838982a4f59a386ec96ac6780c8)


jedidiahkkh:
will likely have to split into two PRs, one for authStore and one for configStore, as I am already at about 600 lines

jedidiahkkh:
updates since last discussion with @sebastianquek,
* added a deleteInStoreInBuckets function to delete keys totally
* replace bucket logic to use string splitting instead, functions now take in strings and return strings instead of Record objects
* updated function JDocs
* made variable names more generic
* move authStore stuff into sub-folder
* had to fix a test at LoginOTPCard that was breaking because of no mocking of SecureStore

jedidiahkkh:
Possible way to break the saving logic:

Currently, we are required to save multiple keys at once. There are two ways to do this, one is to lock the saving logic (the Effect in authStore.tsx:65) and the other is to just run everything without locking.

With locking: no other updates can happen while we are saving the current value. If there are any changes during this time, we will lose them unless we set up some kind of queue for changes.
Without locking: if another update happens while we are saving the current value, the async calls from both might run concurrently and override each other.

Caveat: this requires the two updates to happen really close together, which is unlikely as it means a user needs to add a new campaign/logout successively really quickly.

jedidiahkkh:
@sebastianquek I tested the update logic and found that it doesn't work because when two stateful variables are updated react does 2 separate renders for each update so I can't have both `hasLoadedFromStore` and `authCredentials` update in one atomic render.

I have added a `hasLoadedFromPrimaryStore` state variable that will be updated first and toggle the store saving effect on while keeping `hasLoadedFromStore` false until all store related writes and updates have been completed

jedidiahkkh:
Note: Base on https://docs.expo.io/versions/v42.0.0/sdk/securestore/, SecureStore is not compatible with Android devices running Android 5 and lower. The current minimum version for the app is 5, so it will be affected

jedidiahkkh (commented):
Note: Based on https://docs.expo.io/versions/v42.0.0/sdk/securestore/, SecureStore is not compatible with Android devices running Android 5 and lower. The current minimum version for the app is 5, so it will be affected

jedidiahkkh:
> Note: Based on https://docs.expo.io/versions/v42.0.0/sdk/securestore/, SecureStore is not compatible with Android devices running Android 5 and lower. The current minimum version for the app is 5, so it will be affected

Just tested on emulated Android 5 and it somehow works

src/context/authStore/bucketStorageHelper.ts
```diff
@@ -0,0 +1,120 @@
...
+ * An empty string will be counted as data. This differentiates it from no data.
+ *
+ * @param storageKey key used in storage
+ * @returns the value of the data in storage, null if there was no data found
+ */
+export const readFromStoreInBuckets: (
+  storageKey: string
+) => Promise<string | null> = async (storageKey) => {
+  let bucketNo = 0;
+  let storageData: string | null = null;
```
sebastianquek, line 80: What if we initialise this to an empty string instead? Then we won't need the condition on line 94.

If we want to return null for empty string, we can consolidate when returning, e.g.
```
return storageData.length === 0 ? null : storageData
// Note that this implies line 93 breaks rather than returns
```

src/context/authStore/bucketStorageHelper.ts
```diff
@@ -0,0 +1,120 @@
...
+export const deleteInStoreInBuckets: (
+  storageKey: string,
+  oldValue: string | null
+) => Promise<void> = async (storageKey, oldValue) => {
+  if (oldValue === null) {
+    return;
+  }
+  const bucketCount = Math.max(1, Math.ceil(oldValue.length / SIZE_LIMIT));
+  for (let bucketNo = 0; bucketNo < bucketCount; bucketNo++) {
+    SecureStore.deleteItemAsync(storageKey + "_" + bucketNo);
```
sebastianquek, line 118: Do we want to `await` for this?

src/context/authStore/bucketStorageHelper.ts
```diff
@@ -0,0 +1,120 @@
...
+    );
+
+    if (newValueBucketString !== oldValueBucketString) {
+      if (newValueBucketString === "" && bucketNo !== 0) {
+        // if the new bucket is empty, delete the key
+        // however, if this is the first bucket, user is saving an empty string, so do not delete
+        SecureStore.deleteItemAsync(storageKey + "_" + bucketNo);
+      } else {
+        SecureStore.setItemAsync(
+          storageKey + "_" + bucketNo,
```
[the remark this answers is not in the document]
sebastianquek: Actually, while looking at the other parts, I suggest to move this out to a small utility function to generate the bucket key. So that all functions that require it can use this util rather than knowing about the specific delimiter and format (:

src/context/authStore/authStore.tsx
```diff
@@ -0,0 +1,224 @@
...
+
+  /** Effect to update the store when there are changes to authCredentials */
+  useEffect(() => {
+    if (hasLoadedFromPrimaryStore) {
+      const authCredentialsString = JSON.stringify(authCredentials);
+      const prevAuthCredentialsString = JSON.stringify(prevAuthCredentials);
+      // do a top level check to see if there are any changes
+      if (authCredentialsString !== prevAuthCredentialsString) {
+        // if there are changes, do a check for each bucket
+        saveToStoreInBuckets(
```
sebastianquek, line 90: I noticed we don't currently handle this promise rejecting. What happens if this rejects, e.g. the edge case for when the user has no more space?

I'm thinking we can log a sentry event so that we can monitor when saving to store fails. What do you think? Is there a need for some special handling for such errors?

sebastianquek (changes requested):
Great work!! Have added a couple of comments, I'll look at the tests and authStore in another pass.

src/context/authStore/bucketStorageHelper.ts
```diff
@@ -0,0 +1,120 @@
...
+ * @param oldValue previous value stored
+ * @param bucketSize number of keys per bucket
+ * @param storageKey key to be used for storing
+ */
+export const saveToStoreInBuckets: (
+  storageKey: string,
+  newValue: string,
+  oldValue: string | null
+) => void = (storageKey, newValue, oldValue) => {
+  oldValue = oldValue ?? "";
```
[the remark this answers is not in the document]
jedidiahkkh: I was considering a possible usage pattern where the user does `readFromStoreInBuckets` and then uses that return value as `oldValue` to set a new value (say in a case where the calling code did not previously store the current value), in which case, the return value from `readFromStoreInBuckets` could be `null` if the key had not been set prior.

src/context/authStore/bucketStorageHelper.ts
```diff
@@ -0,0 +1,120 @@
...
+    const oldValueBucketString = oldValue.slice(
+      bucketNo * SIZE_LIMIT,
+      (bucketNo + 1) * SIZE_LIMIT
+    );
+
+    if (newValueBucketString !== oldValueBucketString) {
+      if (newValueBucketString === "" && bucketNo !== 0) {
+        // if the new bucket is empty, delete the key
+        // however, if this is the first bucket, user is saving an empty string, so do not delete
+        SecureStore.deleteItemAsync(storageKey + "_" + bucketNo);
```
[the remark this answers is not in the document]
jedidiahkkh: I am not too sure which is better and whether it will solve the following hypothetical problem I thought of. Although from a reusable code standpoint, it does make sense to `await` for it and leave it up to the caller if they want to block or carry on.

> Possible way to break the saving logic:
> 
> Currently, we are required to save multiple keys at once. There are two ways to do this, one is to lock the saving logic (the Effect in authStore.tsx:65) and the other is to just run everything without locking.
> 
> With locking: no other updates can happen while we are saving the current value. If there are any changes during this time, we will lose them unless we set up some kind of queue for changes.
> Without locking: if another update happens while we are saving the current value, the async calls from both might run concurrently and override each other.
> 
> Caveat: this requires the two updates to happen really close together, which is unlikely as it means a user needs to add a new campaign/logout successively really quickly.

src/context/authStore/bucketStorageHelper.ts
jedidiahkkh: oops missed this out

src/context/authStore/authStore.tsx
jedidiahkkh: would there be any recovery from such an error?

jedidiahkkh: currently we don't await because the function `saveToStoreInBuckets` does not await internally either. Will depend on the above discussion https://github.com/rationally-app/mobile-application/pull/398#discussion_r678119850

src/context/authStore/bucketStorageHelper.ts
```diff
@@ -0,0 +1,120 @@
...
+          newValueBucketString
+        );
+      }
+    }
+  }
+};
+
+/**
+ * Reads data from storage in buckets and returns it. If no data was read, returns null.
+ * An empty string will be counted as data. This differentiates it from no data.
```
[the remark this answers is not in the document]
jedidiahkkh: Don't see an immediate use case currently as we store JSON stringified objects so the shortest string we will store is `{}`. However, since this is agnostic to the type of data we are storing, was considering a use case where the user is just storing normal strings, in which case they could possibly store an empty string. This emulates the same behaviour as `SecureStore.getItemAsync` which returns `null` if the key is not set but would return an empty string if the key was set but with an empty string.

jedidiahkkh:
@sebastianquek git doesn't seem to be detecting the rename for `authStore.tsx`, in the later PR for configs store I move authStore back to it's original location and bucketHelper to the utils because bucketHelper beceoms used by both auth and config store. Should I move that commit over to this pr instead of the next one?

src/context/authStore/bucketStorageHelper.ts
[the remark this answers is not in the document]
sebastianquek: If we look at `saveToStoreInBuckets` independently, a caller will expect it to be synchronous, i.e. once that fn returns, the store would have been updated. I suggest we use `Promise.all` to ensure that all the async calls are resolved before the fn is returned. Note that the use of `Promise.all` immediately rejects if one of the promises fails.

For the useEffect in authStore, what you pointed out about a race condition with `saveToStoreInBuckets` is indeed possible, though unlikely. The locking can look something like this: https://github.com/facebook/react/issues/14326#issuecomment-472043812 However, this complexity isn't worthwhile here as it's unlikely to happen.

src/context/authStore/authStore.tsx
sebastianquek: Yea we may not know the right recovery approach, could be lack of space, could be lack of permissions (?), etc.
To aid with knowing if this happens at all, could you catch the promise rejection and log the error to sentry?

sebastianquek:
@jedidiahkkh yea sure, can move it back, given you know that the next PR will reverse the change anyway.

src/context/authStore/bucketStorageHelper.test.ts
```diff
@@ -0,0 +1,279 @@
...
+    mockSecureDeleteItem.mockReset();
+  });
+
+  describe("saveToStoreInBuckets", () => {
+    it("should split up data into buckets according to the specified size", async () => {
+      expect.assertions(4);
+
+      await saveToStoreInBuckets(
+        testKey,
+        JSON.stringify({
```
sebastianquek, line 33: Can we move this out to a constant, noticed it's used in the following expect statements

src/context/authStore/bucketStorageHelper.test.ts
```diff
@@ -0,0 +1,279 @@
...
+          key1: veryLongStringA,
+          key2: veryLongStringB,
+          key3: veryLongStringC,
+          key4: veryLongStringA,
+          key5: veryLongStringB,
+        }).slice(4096, 6144)
+      );
+    });
+
+    it("should only overwrite buckets that have changes", async () => {
```
sebastianquek, line 78: Would be good to specify which bucket isn't overwritten

src/context/authStore/bucketStorageHelper.ts
```diff
@@ -0,0 +1,129 @@
...
+};
+
+/**
+ * Deletes all buckets from storage. The number of existing buckets is determined by
+ * the length of {@link oldValue}.
+ *
+ * @param storageKey key used in storage
+ * @param oldValue existing value in storage
+ */
+export const deleteInStoreInBuckets: (
```
sebastianquek, line 118: Just noticed that the naming is a little confusing, would `deleteStoreInBuckets` suffice?

sebastianquek: Also, do we need this function? Noticed it isn't being used outside of tests

src/context/authStore/authStore.tsx
jedidiahkkh: since our saving process is non-atomic, if there is an error, it might lead to corrupted data. is deleting the storage a valid recovery approach?

src/context/authStore/bucketStorageHelper.ts
jedidiahkkh: not currently being used in the authStore but just added for completeness of the utility functions

jedidiahkkh:
Have rebased onto master to have updated expo-42 sdk

src/context/authStore.test.tsx
```diff
@@ -17,14 +42,22 @@ const testCampaignKey = "test-campaign";
 
 describe("AuthStoreContextProvider", () => {
   beforeEach(() => {
-    mockGetItem.mockReset();
-    mockSetItem.mockReset();
+    mockGetItem.mockReset().mockName("asyncGetItem");
```
sebastianquek, line 45: Does mockName get cleared on a mockReset?

src/context/authStore.test.tsx
```diff
@@ -52,21 +85,23 @@ describe("AuthStoreContextProvider", () => {
...
     );
 
-    expect(mockGetItem).toHaveBeenCalledTimes(1);
-    expect(mockGetItem).toHaveBeenCalledWith("AUTH_STORE");
+    expect(mockReadBucket).toHaveBeenCalledTimes(1);
+    expect(mockReadBucket).toHaveBeenCalledWith("AUTH_STORE");
     expect(queryByTestId("loaded")).toBeNull();
 
     await waitFor(() => {
+      expect(queryByTestId("loaded")).toHaveTextContent("true");
```
sebastianquek, line 93: Noticed a minor change in order, did you encounter an issue with this expect being after?

src/context/authStore/authStore.tsx
sebastianquek: It is indeed a valid recovery approach, though not a preferred one.

src/context/authStore.tsx
```diff
@@ -41,24 +37,74 @@ export const AuthStoreContext = createContext<AuthStoreContext>({
...
+        // if there are changes, do a check for each bucket
+        saveToStoreInBuckets(
+          AUTH_CREDENTIALS_STORE_KEY,
+          authCredentialsString,
+          prevAuthCredentialsString ?? ""
+        ).catch((reason) => {
+          Sentry.addBreadcrumb({
+            category: "authStore",
+            message: "save failed",
+          });
```
sebastianquek, line 100: Just wondering, is there a reason why you're adding a breadcrumb and not creating a sentry exception?

src/context/authStore.tsx
```diff
@@ -88,89 +134,85 @@ export const AuthStoreContextProvider: FunctionComponent<{
...
-        return true;
-      } else {
-        return false;
+        if (authCredentialsString) {
+          try {
+            const authCredentialsFromStore: AuthCredentialsMap = JSON.parse(
+              authCredentialsString
+            );
+            setAuthCredentialsMap(authCredentialsFromStore);
+            hasUpdatedData = true; // should this be set b4 parsing? error recovery policy
```
sebastianquek, line 162: Can remove the comment here (:

src/context/authStore.test.tsx
jedidiahkkh: Yes, `mockReset` clears everything in the mock function, including implementation, results and the name.

src/context/authStore.test.tsx
jedidiahkkh: not a functional issue but more by convention and also readability of errors. The documentation explains `waitFor` as a function to get an element, so it actually returns whatever the inner function returns and the react native example does it that way with `getByTestId`. The only cosmetic issue with expect being after (inside the waitFor) is that if the `expect` fails, `waitFor` will keep running it until it timesout resulting in an error like `88 assertions instead of 4`

src/context/authStore.tsx
jedidiahkkh: decided to throw it up to the error boundary which will log as an exception
The alternative is to proceed without saving credentials in the storage so everything will be temporary. Is that preferred?

jedidiahkkh: decided to throw it up to the error boundary which will log as an exception
The alternative is to proceed without saving credentials in the storage so everything will be temporary. Is that preferred?

src/context/authStore.tsx
jedidiahkkh: > This is a good point, if the user's existing auth credentials is somehow un-parseable, then they'll be stuck in a loop of migration. I think this is OK because it could have been an error we made. In such a scenario, we can push an update and no data will be lost. If we move this up to before the parsing, then we may accidentally wipe the async storage.

I was thinking about this and realised, even if the data is not lost, once the user adds a new credential, it will be saved to SecureStore. On the next start up of the app, AuthStore will detect that SecureStore has data and wipe the AsyncStorage too. Is that acceptable?

sebastianquek (approved):
Awesome work Jedidiah! 🚢

PR #416, opened 2021-08-10
feat: add logout screen

jedidiahkkh:
[Insufficient session token expiration management (PT-M003)](https://www.notion.so/Insufficient-session-token-expiration-management-PT-M003-f1b6679098c14bd98063013151604539) 

This PR adds a logout spinner screen during which the app contacts the /auth/logout endpoints of all logged in campaigns to invalidate session tokens. Should be taken together with https://github.com/rationally-app/backend-api/pull/858

- [ ] I've kept this PR as small as possible (~600 lines) by splitting it into PRs with manageable chunks of code
- [x] I've added [Chinese translations for i18n setup](https://www.notion.so/Translations-Dev-Guide-8c1da838982a4f59a386ec96ac6780c8)


jedidiahkkh:
are e2e tests required for this?

src/services/auth/index.tsx
```diff
@@ -172,5 +191,95 @@ export const mockValidateOTP = async (
...
+      e.message === "Auth token is not currently valid" ||
+      e.message === oldBackendError + sessionToken
+    ) {
+      // for this case the session token is already invalidated or past the valid date
+      // logoutTodo: is it safe to keep session token there if already past validity date
+
+      Sentry.addBreadcrumb({
+        category: "logout",
+        message: JSON.stringify({
+          status: "success despite error",
```
szenius, line 238: Wait what is success despite error I'm confused 😅

src/services/auth/index.tsx
```diff
@@ -172,5 +191,95 @@ export const mockValidateOTP = async (
...
+    }
+  }
+};
+
+export const mockCallLogout = async (
+  _sessionToken: string,
+  _operatorToken: string,
+  _endpoint: string
+): Promise<void> => {
+  // throw new LogoutError("error logging out");
```
szenius, line 280: Is this intended?

package.json
```diff
@@ -6,7 +6,7 @@
     "test": "jest",
     "test:coverage": "npm run test -- --coverage",
     "test:watch": "jest --watch",
-    "storybook": "APP_BUILD_VERSION=1 APP_BINARY_VERSION=dev START_STORYBOOK=true expo start",
+    "storybook": "APP_BUILD_VERSION=1 APP_BINARY_VERSION=dev START_STORYBOOK=true MOCK=true expo start",
```
szenius, line 9: Why do we need the extra MOCK?

szenius (commented):
I tried logging out, but kept getting network error, not sure if I missed anything in my setup?

image

szenius:
Oic, this requires the backend PR right. Can you state so on your PR description + indicate on your Notion story, so the deployment folks know there is a dependency?

src/components/Logout/LogoutScreen.tsx
```diff
@@ -0,0 +1,136 @@
...
+            removeCampaignConfig(key);
+          }
+        )
+      );
+      const errorResults = results.filter(
+        (result) => result.status === "rejected"
+      );
+      if (errorResults.length > 0) {
+        const error: Error = (errorResults[0] as PromiseRejectedResult).reason;
+        if (error instanceof LogoutError) {
```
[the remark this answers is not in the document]
jedidiahkkh: it covers `OperatorLogoutError`s as well. Additionally, it should catch any other error thrown from the backend that is unknown for whatever reason. (e.g. their messages don't match anything)

src/components/Logout/LogoutScreen.tsx
```diff
@@ -0,0 +1,136 @@
...
+    top: 0,
+    left: 0,
+    right: 0,
+    bottom: 0,
+    alignItems: "center",
+    justifyContent: "center",
+  },
+});
+
+// hack because react-native does not have Promise.allSettled
```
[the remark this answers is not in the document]
jedidiahkkh: ok, i'll write tests for it in that case as well

src/components/Logout/LogoutScreen.tsx
```diff
@@ -0,0 +1,136 @@
...
+  } = useContext(AuthStoreContext);
+  const { removeCampaignConfig, clearCampaignConfigs } = useContext(
+    CampaignConfigsStoreContext
+  );
+  const { setConfigValue } = useContext(ConfigContext);
+  const setMessageContent = useContext(ImportantMessageSetterContext);
+  const { showErrorAlert } = useContext(AlertModalContext);
+
+  const [, setState] = useState();
+  useEffect(() => {
```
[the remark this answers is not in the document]
jedidiahkkh: i was thinking because the stuff here should only run on the logout screen so it shouldn't be exposed to the other functions call `useLogout` merely to trigger the logout process.

Actually now that you point this out, `useLogout` isn't really a hook anymore, should we instead just change it to a normal function?

src/hooks/useValidateExpiry.tsx
```diff
@@ -46,7 +46,7 @@ export const useValidateExpiry = (
   const validate = useCallback(async (): Promise<void> => {
     clearTimeout(timeout);
 
-    if (!expiry || isLoggingOut) {
+    if (!expiry) {
```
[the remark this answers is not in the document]
jedidiahkkh: it does nothing actually to the rest of the code. No one calls `useValidateExpiry` anymore, it's like a standalone, unused hook.

Within useValidateExpiry, the effect is that there will be a bug if the user was already in the process of logging out when `validate` is called. However, I am of the opinion that we can just delete the whole file.

src/services/auth/index.tsx
```diff
@@ -172,5 +191,95 @@ export const mockValidateOTP = async (
...
+  } catch (e) {
+    if (e instanceof NetworkError) {
+      throw e;
+    } else if (
+      e instanceof SessionError ||
+      e.message === "Auth token is not currently valid" ||
+      e.message === oldBackendError + sessionToken
+    ) {
+      // for this case the session token is already invalidated or past the valid date
+      // logoutTodo: is it safe to keep session token there if already past validity date
```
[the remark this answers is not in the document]
jedidiahkkh: ok! It was more for me to differentiate TODOs that were related to this PR specifically, but I can rename it to something like `TODO: <logout>` that will still be recognised by extensions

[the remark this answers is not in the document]
jedidiahkkh: If a sessionToken on the database is already past the valid date, the logout handler on the backend side will fail. However, I have written the FE to just discard the sessionToken on the FE side since the sessionToken is expired on BE and unusable anyway.

src/services/auth/index.tsx
jedidiahkkh: Error is thrown by the backend, but we can still consider it as a successful logout.
E.g. if the sessionToken is already non-existent on the BE, it will throw an error. In this case the user does not need to do any further action (because the sessionToken already doesn't work anyway), so we just silently discard the sessionToken on the FE side.

src/services/auth/index.tsx
jedidiahkkh: yep it's deliberately commented out, it's for if a failed Logout needs to be tested during mocking.

package.json
jedidiahkkh: so that the logout screen can load on the Storybook without having the logout handlers run in the background and fail

src/components/Logout/LogoutScreen.tsx
```diff
@@ -0,0 +1,136 @@
...
+        )
+      );
+      const errorResults = results.filter(
+        (result) => result.status === "rejected"
+      );
+      if (errorResults.length > 0) {
+        const error: Error = (errorResults[0] as PromiseRejectedResult).reason;
+        if (error instanceof LogoutError) {
+          showErrorAlert(error);
+          navigation.navigate("CampaignLocationsScreen", { autoLoad: false });
```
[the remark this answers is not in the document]
jedidiahkkh: i'm not too sure, I'll try test this out. My guess is that navigation parameters don't persist, but will confirm.

src/components/Logout/LogoutScreen.tsx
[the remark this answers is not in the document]
szenius: Hmm. If not being used by other functions, can remove the hook and keep in the logout component bah!

src/hooks/useValidateExpiry.tsx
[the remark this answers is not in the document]
szenius: If we are certain it's not being used, then please do remove it!

src/services/auth/index.tsx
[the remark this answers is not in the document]
szenius: I thought the finding says to remove it from the backend?

src/services/auth/index.tsx
szenius: Can we describe it better in the status message / a comment?

src/services/auth/index.tsx
szenius: Can we leave a comment above to explain? Something like 

```
// Uncomment to test throwing an error
// throw new LogoutError("error logging out");
```

package.json
szenius: How come we cannot use `APP_BINARY_VERSION`? How are the rest of the components running on storybook without this MOCK attribute?

src/services/auth/index.tsx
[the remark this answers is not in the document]
jedidiahkkh: The finding was about session tokens that are still able to be used even though the user has logged out from the app. In this case, because the validity of the session token has expired, we will not be able to use the session token as the authenticatedHandler will block at the middleware, so an attacker also can't use the session token. The side effect is also that we also can't use the session token to call the handler to log out.

package.json
jedidiahkkh: Because the rest of the screens don't have any calls to services. They go through the `AuthStoreContextProvider` and `ConfigStoreContextProvider` which are not used in the storybook (which uses custom `AuthStoreContext.Provider`s). There's only one other screen that calls services which is `CampaignInitialisationScreen` however that's not shown in storybook

src/components/Logout/LogoutScreen.tsx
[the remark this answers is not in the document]
jedidiahkkh: Yep, navigation parameters do not persist. Have tested the following flow

1. Login to only one campaign
2. Logout -> error
3. Campaign Location Screen does not autoload
4. Select campaign
5. Logout -> success
6. Login to one campaign
7. Go to campaign location screen
8. Campaign Location Screen autoloads

src/services/auth/index.tsx
```diff
@@ -172,5 +191,95 @@ export const mockValidateOTP = async (
   };
 };
 
+/** Error message returned by endpoints without a /auth/logout handler */
+const oldBackendError =
+  "Authorization header requires 'Credential' parameter. Authorization header requires 'Signature' parameter. Authorization header requires 'SignedHeaders' parameter. Authorization header requires existence of either a 'X-Amz-Date' or a 'Date' header. Authorization=";
```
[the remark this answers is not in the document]
jedidiahkkh: what do you mean by a class? Like a separate file that just stores error messages from the backend?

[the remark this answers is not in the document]
szenius: Error classes. I think we have others in this codebase too, can ref

src/components/Layout/DrawerNavigation.tsx
```diff
@@ -19,6 +18,7 @@ import { useDrawerContext, DrawerButton } from "../../context/drawer";
 import Constants from "expo-constants";
 import { AlertModalContext, CONFIRMATION_MESSAGE } from "../../context/alert";
 import { useTranslate } from "../../hooks/useTranslate/useTranslate";
+import { logout } from "../Logout/LogoutScreen";
```
szenius, line 21: Actually if it’s only being used here, maybe the function should only be in this file instead?

src/components/Logout/LogoutScreen.test.tsx
```diff
@@ -0,0 +1,298 @@
...
+
+const defaultCampaignConfigStoreContextValue = {
+  allCampaignConfigs: {},
+  clearCampaignConfigs: jest.fn(),
+  hasLoadedFromStore: true,
+  removeCampaignConfig: jest.fn(),
+  setCampaignConfig: jest.fn(),
+};
+
+const defaultConfigContextProviderValue = {
```
szenius, line 87: Try to create these variables this way instead

```
describe(‘test’, () => {
  let var1;
  let var2;
  let …

  beforeAll(() => {
    var1 = …;
    var2 = …;
    …
  });
});
```

The reason is that all the statements outside of the before block will be executed regardless of whether this test should be executed. I.e. if I were just running a single test file A, all the variables initialised outside of the `beforeAll` block in test file B will also be run. This unnecessarily increases the overhead of running tests.

src/components/Logout/LogoutScreen.test.tsx
```diff
@@ -0,0 +1,298 @@
...
+
+    await waitFor(() => expect(mockNavigate).toHaveBeenCalledTimes(1));
+
+    expect(mockCallLogout).toHaveBeenCalledTimes(3);
+    expect(mockCallLogout).toHaveBeenCalledWith("sesA", "optA", "endA");
+    expect(mockCallLogout).toHaveBeenCalledWith("sesB", "optB", "endB");
+    expect(mockCallLogout).toHaveBeenCalledWith("sesC", "optC", "endC");
+  });
+
+  it("should logout from campaigns with no error and vice versa", async () => {
```
szenius, line 124: This description can probably be improved? It’s very ambiguous now. Maybe 

```
should logout from all other campaigns even when one of them throws
```

Or better,

```
describe(“when logging out of one of the campaigns throws error”, () => {
  it(“should log out of all other campaigns”, () => {…});
});
```

but whether to do the first or second, I’ll leave it up to you. Better to have consistency within the file too.

src/components/Logout/LogoutScreen.test.tsx
```diff
@@ -0,0 +1,298 @@
...
+    expect(mockRemoveAuthCredentials).toHaveBeenCalledWith(keyB);
+    expect(mockRemoveAuthCredentials).toHaveBeenCalledWith(keyC);
+
+    expect(mockRemoveCampaignConfig).toHaveBeenCalledTimes(2);
+    expect(mockRemoveCampaignConfig).not.toHaveBeenCalledWith(keyA);
+    expect(mockRemoveCampaignConfig).toHaveBeenCalledWith(keyB);
+    expect(mockRemoveCampaignConfig).toHaveBeenCalledWith(keyC);
+  });
+
+  it("should clear all credentials and configs and move to login screen upon successful logout", async () => {
```
szenius, line 166: Can this be done under the first `it` above?

src/components/Logout/LogoutScreen.test.tsx
```diff
@@ -0,0 +1,298 @@
...
+    );
+
+    await waitFor(() => expect(mockNavigate).toHaveBeenCalledTimes(1));
+
+    expect(mockNavigate).toHaveBeenCalledWith("LoginScreen");
+    expect(mockClearAuthCredentials).toHaveBeenCalledTimes(1);
+    expect(mockClearCampaignConfigs).toHaveBeenCalledTimes(1);
+  });
+
+  it("should not clear all data and should move to campaign selection screen upon unsuccessful logout", async () => {
```
szenius, line 197: Hmm is this when there is 1 single campaign or multiple campaigns? Is it an extension of the second `it` block above?

szenius: If you find that you are setting up for the test case the same way in many tests, you can reorganise your tests into this structure

```
describe(‘name of tests’, () => {
  describe(‘when there are multiple campaigns’, () => {
      describe(‘when logout for all campaigns is successful’, () => {
        beforeAll(() => { 
          // set up for test here once! 
        });

        it(‘should clear all data’, () => { 
          // …
      });

      describe(‘when logout for one of the campaigns failed’, () => {
        beforeAll(() => { 
          // set up for test here once! 
        });
        it(‘should do something about state’, () => {…});
        it(‘should show some alert’, () => {…});
      });
  });

  describe(‘when there is 1 campaign’, () => {…});
});
```

This way it’s easier to understand your tests because they are structured by scenarios. And you don’t have so much duplicated code trying to set up the tests the same way every time. Can consider~

szenius (commented):
Some comments on structuring your tests.

I retested the logout itself and it looks good! What was the issue?

src/services/auth/index.tsx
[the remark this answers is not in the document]
jedidiahkkh: oh, this one isn't an error that we throw though, it's an error message that we receive from the backend. I actl have no idea how this error message gets generated on the backend, but based on what I tested this is the error that we get if we call `auth/logout` on a backend that hasn't created the logout endpoint so it's probably an amazon autogenerated message.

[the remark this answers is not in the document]
szenius: Same thing though, there are error classes we define in the frontend codebase based on what the backend throws. Just like all the login error classes?

src/services/auth/index.tsx
```diff
@@ -172,5 +191,97 @@ export const mockValidateOTP = async (
...
+  endpoint: string
+): Promise<void> => {
+  const payload = { sessionToken, operatorToken };
+
+  try {
+    await fetchWithValidator(LogoutResponse, `${endpoint}/auth/logout`, {
+      method: "POST",
+      headers: {
+        Authorization: sessionToken,
+        "Content-Type": "application/json",
```
jedidiahkkh, line 210: Note for future reference:
Required this line here as without this header the `fetch` call would throw a `TypeError: Network service unavailable` without even attempting to call the backend. No packet would be sent out.

src/services/auth/index.tsx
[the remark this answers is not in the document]
jedidiahkkh: Ah, not sure if I get what you mean but the error class is to replicate what is at the backend? I think the difference for this one is that I am only using the string constant at line 231 for message checking and I don't throw an error again, so if we define an error class for this we also wouldn't use it as we don't rethrow the error up.

src/components/Logout/LogoutScreen.test.tsx
jedidiahkkh: This should be true regardless of whether or not there is one campaign or many. Currently, the many campaign case is the default and there's no special treatment for that so I didn't test it. The special case is when there is only one campaign left after logout which is what is being tested here (`campaign A` is the one that is remaining here)

src/components/Logout/LogoutScreen.test.tsx
jedidiahkkh: i refactored it slightly differently to sort of match the structure you gave [here](https://github.com/rationally-app/mobile-application/pull/416#discussion_r694030520)

szenius (approved):
LGTM. Thank you for the hard work on this :)

szenius (approved):
No changes after rebase, LGTM

PR #419, opened 2021-08-13
feat: use securestore for config

jedidiahkkh:
Plaintext token storage in unencrypted database files (PT-M004 

Changes storage of secrets (operator token, session token, endpoints) to SecureStore which is encrypted instead of Asyncstorage that is unencrypted.

This PR is for the changes to ConfigStore components
* ConfigStoreProvider
with minor docs change to AuthStore components
* AuthStoreProvider
Changes for authStore were in a separate [PR](https://github.com/rationally-app/mobile-application/pull/398) 

- [ ] (no UI change) I've added [accessibility IDs](https://www.notion.so/dc5f4ce910f7431c84344ac79344e9f5?v=d487366816834dd4ab8dc12e0b5928f3) to my components
- [ ] (no UI change) I've added [Chinese translations for i18n setup](https://www.notion.so/Translations-Dev-Guide-8c1da838982a4f59a386ec96ac6780c8)


src/context/campaignConfigsStore.tsx
```diff
@@ -107,27 +134,80 @@ export const CampaignConfigsStoreContextProvider: FunctionComponent = ({
...
+      }
+      if (hasUpdatedData) {
+        await AsyncStorage.removeItem(CAMPAIGN_CONFIGS_STORE_KEY);
       }
+
+      return hasUpdatedData;
+    };
+
+    const loadCampaignConfigFromStore = async (): Promise<void> => {
+      let newStorageHasData = false;
```
andrehadianto, line 176: instead of having a variable as toggle, what do you think if we use `campaignConfigsString` instead?
```suggestion
      let campaignConfigsString: WhateverTheTypeIs;
```

src/context/campaignConfigsStore.tsx
```diff
@@ -107,27 +134,80 @@ export const CampaignConfigsStoreContextProvider: FunctionComponent = ({
...
+          }));
+          newStorageHasData = true;
+        }
       } catch (e) {
         setState(() => {
           throw new CampaignConfigError(e);
         });
       }
+      setHasLoadedFromPrimaryStore(true);
+      const migrated = await migrateOldConfigFromStore(newStorageHasData);
```
andrehadianto, line 195: then you can pass `await migrateOldConfigFromStore(!!campaignConfigsString);` or if you want to be cleaner, create the `newStorageHasData` variable just before this instead.

src/context/campaignConfigsStore.tsx
```diff
@@ -107,27 +134,80 @@ export const CampaignConfigsStoreContextProvider: FunctionComponent = ({
...
+      if (!newStorageHasData) {
+        // migration was attempted
+        if (migrated) {
+          Sentry.addBreadcrumb({
```
andrehadianto, line 199: and with that, do you think we can move these `Sentry` inside  `migrateOldConfigFromStore` instead?

src/context/campaignConfigsStore.tsx
```diff
@@ -107,27 +134,80 @@ export const CampaignConfigsStoreContextProvider: FunctionComponent = ({
...
+        const configString = await AsyncStorage.getItem(
+          CAMPAIGN_CONFIGS_STORE_KEY
+        );
+        if (configString) {
+          try {
+            const configStringFromStore: CampaignConfigsMap = JSON.parse(
+              configString
+            );
+            setAllConfigs(configStringFromStore);
+            hasUpdatedData = true;
```
andrehadianto, line 160: since i notice that `hasUpdatedData` will only remain false if this `try catch` fail.

so maybe we can put the `Sentry` right after the catch and right before `return hasUpdatedData`

wdyt?

andrehadianto (changes requested):
some comments!

src/context/campaignConfigsStore.tsx
jedidiahkkh: `hasUpdatedData` is not guaranteed to be `true` at the end of `migrateOldConfigFromStore` even if there is no error because there is a chance that AsyncStorage does not have any data as well (`configString === null`), in which case, there is nothing to migrate and the function will return false

jedidiahkkh:
Since the code structure in ConfigStore is almost identical to AuthStore (#398), shall I also replicate any changes here over on the AuthStore side as well?

src/context/campaignConfigsStore.tsx
jedidiahkkh: would it potentially be a problem if `campaignConfigsString` is an empty string `""`? Because then it would be a falsy value.
Although by right if it is an empty string the JSON parse would throw an error already, but then the issue would be with the conditions being slightly different (line 195 is non-null and line 209 is truthy). Do you think this is an issue?

Or we can also do `await migrateOldConfigFromStore(campaignConfigsString !== null);`

PR #508, opened 2021-12-23
refactor: use c13n instead of hardcoded copy changes for GW theme

dalsontws:
This PR adds... 
- use of c13n for previously hardcoded app text
- toggle these text based on SSM params `en` and `zh` if they exist

### DIFF
--- a/src/components/CustomerDetails/InputIdSection.tsx
+++ b/src/components/CustomerDetails/InputIdSection.tsx
@@ -7,8 +7,6 @@ import { AppText } from "../Layout/AppText";
 import { SecondaryButton } from "../Layout/Buttons/SecondaryButton";
 import { size, color, fontSize } from "../../common/styles";
 import { useTranslate } from "../../hooks/useTranslate/useTranslate";
-import { useTheme } from "../../context/theme";
-import { GOVWALLET_THEME_NAME } from "../../common/styles/themes";
 
 const styles = StyleSheet.create({
   scanButtonWrapper: {
@@ -57,8 +55,8 @@ export const InputIdSection: FunctionComponent<InputIdSection> = ({
   submitId,
   keyboardType,
 }) => {
-  const { theme } = useTheme();
-  const { i18nt } = useTranslate();
+  const { i18nt, c13nt } = useTranslate();
+  const checkKey = "check";
   return (
     <>
       <View style={styles.scanButtonWrapper}>
@@ -94,9 +92,9 @@ export const InputIdSection: FunctionComponent<InputIdSection> = ({
         </View>
         <SecondaryButton
           text={
-            theme.name === GOVWALLET_THEME_NAME
-              ? i18nt("collectCustomerDetailsScreen", "record")
-              : i18nt("collectCustomerDetailsScreen", "check")
+            c13nt(checkKey) !== checkKey
+              ? c13nt(checkKey)
+              : i18nt("collectCustomerDetailsScreen", checkKey)
           }
           onPress={submitId}
           accessibilityLabel="identity-details-check-button"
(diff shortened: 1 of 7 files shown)


src/components/CustomerQuota/NoQuota/TransactionTitle.tsx
```diff
@@ -7,32 +7,24 @@ import {
...
-              {
-                dateTime: formatDateTime(transactionTime),
-                today,
-              }
-            )
-          : i18nt("checkoutSuccessScreen", "limitReachedDate", undefined, {
+        {c13nt(limitReachedDateKey) !== limitReachedDateKey
+          ? c13nt(limitReachedDateKey)
+              .replace("${dateTime}", formatDateTime(transactionTime))
+              .replace("${today}", today)
```
dalsontws, line 26: Not sure if this is the best approach, wanted to use [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) (i.e. `${dateTime}` and assign the value) but wanted to keep it consistent with the rest of the text, hence using this format instead. Do comment if there are ways this can be done better!

src/components/CustomerDetails/InputIdSection.tsx
```diff
@@ -94,9 +92,9 @@ export const InputIdSection: FunctionComponent<InputIdSection> = ({
         </View>
         <SecondaryButton
           text={
-            theme.name === GOVWALLET_THEME_NAME
-              ? i18nt("collectCustomerDetailsScreen", "record")
-              : i18nt("collectCustomerDetailsScreen", "check")
+            c13nt(checkKey) !== checkKey
+              ? c13nt(checkKey)
+              : i18nt("collectCustomerDetailsScreen", checkKey)
```
szenius, line 97: Hmm I'm assuming you're trying to default to `i18nt` when the `c13n` object doesn't contain the translation.

Actually `c13nt`'s 3rd input param takes in a default value. Can see some of the other existing implementations for this. So no need to ownself do ternary check liao

Also how we intended to use c13n was to have the actual string to be displayed as the key and its translation as the value. So something like

```
// c13n/en
{
  "This is a message to be displayed to users": "这是用户会看到的讯息"
}
```

Not sure your implementation follows this pattern?

szenius: Discussed with Dalson on Slack. Here's the summary

* Dalson's idea was that for the same location of any copy, both govwallet and sally themed campaigns should use the same key in c13n
* It makes a lot of sense to use the same key. But perhaps the key can still default to the copy for sally themed campaigns^, so that the defaulting to key in `c13nt` function can continue to do its job
* Then in this case we also don't have random key strings lying around liao

What do you think?

^ Example
```
// for sally themed
{
  "Check": "Check"
}


// for govwallet themed
{
  "Check": "Record"
}
```

dalsontws: thanks for the summary! I've made the change to adhere to the example above, with the GovWallet theme stored in the SSM c13n params (`en` & `zh`)!

src/components/CustomerQuota/NoQuota/TransactionTitle.tsx
```diff
@@ -13,21 +13,23 @@ export const DistantTransactionTitle: FunctionComponent<{
...
+        {`${c13nt(
+          "limitReachedDate",
+          undefined,
+          i18nt("checkoutSuccessScreen", "limitReachedDate", undefined, {
+            dateTime: formatDateTime(transactionTime),
+            today,
+          })
+        )}`
+          .replace("%{dateTime}", formatDateTime(transactionTime))
+          .replace("%{today}", today)}
```
szenius, line 32: Hmm maybe can add a TODO here to move this functionality into c13nt function, so in the future it offers this parameter feature too~ Same for line 58

szenius (commented):
one small comment!

src/components/CustomerQuota/NoQuota/TransactionTitle.tsx
dalsontws: good call, added [here](https://github.com/rationally-app/mobile-application/pull/508/commits/af4ff260f27365aa0f6dcde211f66beeefa9e360)!

szenius (approved):
LGTM!!

PR #512, opened 2021-12-29
fix: De-sync between passport number field and what is submitted

yao-peijun:
[Notion link](https://www.notion.so/sally-wallet/De-sync-between-passport-number-field-and-what-is-submitted-4277f5c27bae45879369ad3ea99e813a)

This PR fix de-sync between passport number field and what is submitted.

- [X] I've kept this PR as small as possible (~600 lines) by splitting it into PRs with manageable chunks of code
- [X] I've requested reviews from 1-2 reviewers
- [ ] I've added [accessibility IDs](https://www.notion.so/dc5f4ce910f7431c84344ac79344e9f5?v=d487366816834dd4ab8dc12e0b5928f3) to my components
- [ ] I've added/updated unit/integration tests
- [ ] I've added/updated [e2e tests](https://www.notion.so/e2e-Documentation-a096d3e0bf75485e85ad692af8371ef3) for at least the happy flow
- [ ] I've added [Chinese translations for i18n setup](https://www.notion.so/Translations-Dev-Guide-8c1da838982a4f59a386ec96ac6780c8)
- [ ] I've added documentation in README/[Notion](https://www.notion.so/82b92fb1007640328dab9582c0a3694e?v=3b6ce48202cf40ad8450553799b13146)

PR #529, opened 2022-01-12
fix: Revert "fix: De-sync between passport number field and what is submitted"

dalsontws:
Reverts rationally-app/mobile-application#512


andrehadianto (approved):
approvedd

ilhaoni (approved):
WOHOO

dalsontws:
We don't need this anymore since the story was tested and demo-ed.

PR #537, opened 2022-01-19
feat: support parsing NETS payment QRs

yongjiajie:
[Notion link](https://www.notion.so/sally-wallet/Parse-Validate-and-Save-into-s3-Leverage-Sally-app-to-capture-payment-QR-for-merchant-ringfencing-a4f8b184c0114285b4fd2c4bf44f427d) 

This PR adds... 

**Payment QR**
- Errors when parsing payment QR
- Renamed `paymentQrValidate` to `isValidPaymentQR` for clarity
- Handle errors when validating payment QR identifiers, so that we can show appropriate error dialogs as per story

**Error Dialogs**
- Add `i18n` translations for payment QR related error dialogs
- Added mappings from error to error dialog content

**Note**: `zh` translations are not needed for new error dialogs, hence the `title`s are not translated. However, since the `body`s were translated before, the `zh` equivalents are used.

**Miscellaneous**
- Bump `@rationally-app/payment-qr-parser` to latest version (`1.7.0`)
- Removed unnecessary `waitFor` in `campaignConfigsStore.test.tsx` that were throwing errors during testing
- Extended timeout for long running tests in `campaignConfigsStore.test.tsx` and `authStore.test.tsx`


src/common/i18n/translations/zh.ts
```diff
@@ -440,5 +440,20 @@ export const zh: Translations = {
...
+    missingInfoInPaymentQR: {
+      title: "Missing info in payment QR",
+      body: "请输入或扫描有效的编号。",
+      primaryActionText: "确定",
+    },
+    unsupportedPaymentMethodInPaymentQR: {
+      title: "Unsupported payment method",
+      body: "请输入或扫描有效的编号。",
+      primaryActionText: "确定",
+    },
```
ilhaoni, line 457: 😂  Let's check with Eida? I'm trying to translate, but I'm having trouble too 😂

ilhaoni: I'm thinking 
deformedPaymentQR: 支付QR码形状变形
missingInfoInPaymentQR: 支付QR码缺少信息
unsupportedPaymentMethodInPaymentQR: 支付QR码不在名单中

src/components/CustomerQuota/CustomerQuotaScreen.tsx
```diff
@@ -180,13 +185,19 @@ export const CustomerQuotaScreen: FunctionComponent<CustomerQuotaProps> = ({
...
 
     if (cartError) {
-      if (cartError instanceof SessionError) {
-        clearCartError();
-        expireSession();
-        showErrorAlert(cartError, () => {
-          navigation.navigate("CampaignLocationsScreen");
-        });
-        return;
+      switch (true) {
```
ilhaoni, line 188: should we have a default case? 🤔  would the code ever go there?

src/context/authStore.test.tsx
```diff
@@ -146,7 +146,7 @@ describe("AuthStoreContextProvider", () => {
     await waitFor(() => {
       expect(mockCaptureException).toHaveBeenCalledTimes(1);
     });
-  });
+  }, 60000);
```
ilhaoni, line 149: is this the timeout you set to fix the test 😆

ilhaoni (commented):
overall LGTM :D 

just some questions~

src/context/authStore.test.tsx
yongjiajie: Yeah, this test and another test in `campaignConfigsStore.test.tsx`. Both of them are performing the same test that waits for `Sentry` to capture the exception 🤔

src/common/i18n/translations/zh.ts
yongjiajie: Will check in with Eida!

src/components/CustomerQuota/CustomerQuotaScreen.tsx
ilhaoni: thats true... hmm 🤔  currently if we don't put in the default case, whats the current behaviour? 

I think putting in a default case might not work, cause it might skip the if checks below >< I went to reread everything HAHAHA

src/context/authStore.test.tsx
ilhaoni: I see I see! Hmm... does this part makes sense ah? Or should we mock that Sentry returns ok, since this test seems to be focused on testing other logic. What do you think? 🤔

src/common/i18n/translations/zh.ts
yongjiajie: Checked in with Eida, she mentioned that we will not need the Chinese translations, since the users will be English speaking!

src/components/CustomerQuota/CustomerQuotaScreen.tsx
yongjiajie: After reading through the code again, I think putting no `default` case makes more sense.

Without a `default` case, if `cartError` does not match any of the `Error` types, execution will proceed to evaluate the `if` block below!

We can put a `default` case with nothing inside, but that is just a waste of space 😂

ilhaoni: Yeah.... HAHAHA. Okok we have the same understanding of how the code runs HAHAHA. Ok then I guess let's leave it as is HAHAHA

src/context/authStore.test.tsx
ilhaoni: @sebastianquek what do you think? Good to remove the tests? :P

src/components/CustomerQuota/CustomerQuotaScreen.tsx
yongjiajie: I wrote a snippet and it works as intended:

- If `cartError` matches any of the errors in the switch-case, it will go into that branch.
- Else, the function will continue to evaluate the if block below

ilhaoni: Nice, thanks for checking! ^^

src/components/CustomerQuota/CustomerQuotaScreen.tsx
```diff
@@ -34,6 +34,11 @@ import { SessionError } from "../../services/helpers";
 import { useQuota } from "../../hooks/useQuota/useQuota";
 import { AuthStoreContext } from "../../context/authStore";
 import { useTranslate } from "../../hooks/useTranslate/useTranslate";
+import {
+  PaymentQRDeformedError,
+  PaymentQRMissingInfoError,
+  PaymentQRUnsupportedError,
+} from "../../utils/paymentQrValidation";
```
dalsontws, line 41: can we import these errors from the `payment-qr-parser` lib instead?

src/utils/validateIdentifierInputs.test.ts
```diff
@@ -307,3 +313,88 @@ describe("validateIdentifierInputs", () => {
...
+          validationRegex: "[\\s\\S]*",
+          textInputType: "PAYMENT_QR",
+          value:
+            "00020101021102164761360000000*1704155123456789123451110123456789012153123456789012341531250003440001034450003445311000126330015SG.COM.DASH.WWW0110000005550127670014A00000076200010120COM.LQDPALLIANCE.WWW021512345678901234503020028660011SG.COM.OCBC0147OCBCP2P629A358D-ECE7-4554-AD56-EBD12D84CA7E4F7329500006SG.EZI013603600006-76bb-4a5a-aa1a-fbcb64d6ecf530850013SG.COM.EZLINK01201234567890123456-1230204SGQR0324A123456,B123456,C12345670404A23X31260008COM.GRAB0110A93FO3230Q32390007COM.DBS011012345678900210123456789033900011SG.COM.NETS01230201401832831128823590002150001118703240000308885872010901199084E5DC3D834430017COM.QQ.WEIXIN.PAY011012345678900204123435660010SG.COM.UOB014845D233507F5E8C306E3871A4E9FACA601A80C114B5645E5D36840009SG.PAYNOW010120216+621234567890123030100408209912310525123456789012345678901234537270009SG.AIRPAY0110A11BC0000X51860007SG.SGQR0112180307510317020701.0003030608100604020205031380609Counter010708201804075204581453037025802SG5916FOOD XYZ PTE LTD6009SINGAPORE6106081006626001081234567R0607876543050330015SG.COM.DASH.WWW0110000005550263045E8D",
```
dalsontws, line 328: can we try using the `testConstants` from the `payment-qr-parser` lib?

src/components/CustomerQuota/CustomerQuotaScreen.tsx
yongjiajie: Yes, we can, have switched to using the errors from the library instead in [6b0627a](https://github.com/rationally-app/mobile-application/pull/537/commits/6b0627af0dfecf2d23917e128146da46b770dc95)!

src/utils/validateIdentifierInputs.test.ts
yongjiajie: Have changed to use the test constants in the library in [7b3a692](https://github.com/rationally-app/mobile-application/pull/537/commits/7b3a692c4d8788a42de5d70f62fc6a946d25975d)!

src/context/authStore.test.tsx
sebastianquek: The purpose of this test is to ensure that when there's a malformed config, `JSON.parse` will throw an error that should be handled appropriately. Testing the outcome that Sentry is triggered is one possible way. There are other possible ways such as checking the `ErrorBoundaryContent`, but it's out of scope for this story.

So given the 60s works, let's keep it this way and look closer into this issue separately?
Have added a card here: https://www.notion.so/sally-wallet/Investigate-if-timeout-of-60s-on-mobile-app-tests-can-be-removed-a322a3d719ce4a55aaf4aa110a0f7bfd

yongjiajie: I am okay with this, we can tackle this as a chore in the future!

dalsontws (approved):
LGTM!🎉

src/context/authStore.test.tsx
ilhaoni: Sounds good~ thanks Seb! ^^

ilhaoni (approved):
LGTM! :D

PR #539, opened 2022-01-20
fix: add sessionExpired error msg and map it to sessionError in alertContext

yao-peijun:
Given someone else uses your QR to login, when you try to enter an NRIC to check for quota, you will see an expired modal. This can be confusing because it's not expired, but rather, the code has already been used by someone else. [Notion link](https://www.notion.so/sally-wallet/Session-errors-should-be-broken-down-further-bfb7d69e5a9747ce8ef8c3383733843c)

This PR adds sessionExpired [error msg](https://www.notion.so/sally-wallet/Error-Message-c8d8b16fec154e738f7274026616b0a7) and map it to sessionError in alert. 

- [X] I've kept this PR as small as possible (~600 lines) by splitting it into PRs with manageable chunks of code
- [X] I've requested reviews from 1-2 reviewers
- [X] I've added [Chinese translations for i18n setup](https://www.notion.so/Translations-Dev-Guide-8c1da838982a4f59a386ec96ac6780c8)

### DIFF
--- a/src/components/DailyStatistics/DailyStatisticsScreen.tsx
+++ b/src/components/DailyStatistics/DailyStatisticsScreen.tsx
@@ -3,12 +3,14 @@ import React, {
   useEffect,
   useContext,
   useState,
+  useCallback,
 } from "react";
 import { View, StyleSheet } from "react-native";
 import { size, fontSize } from "../../common/styles";
 import { TopBackground } from "../Layout/TopBackground";
 import { Credits } from "../Credits";
 import { AuthContext } from "../../context/auth";
+import { AuthStoreContext } from "../../context/authStore";
 import { useConfigContext } from "../../context/config";
 import { withNavigationFocus } from "react-navigation";
 import { TitleStatistic } from "./TitleStatistic";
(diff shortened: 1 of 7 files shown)


dalsontws (commented):
looks good! One comment: not specifically stated of this story, but noticed that when a user with an expired session clicks on the `Go to statistics` button, we will throw a general `StatisticsError` instead of the `SessionError`, which should also be the case. Can you help to add the catching of `SessionError` when fetching stats as well?

src/components/DailyStatistics/DailyStatisticsScreen.tsx
```diff
@@ -94,9 +94,7 @@ const DailyStatisticsScreen: FunctionComponent<NavigationProps> = ({
 
   useEffect(() => {
     if (error) {
-      showErrorAlert(new Error(ERROR_MESSAGE.SERVER_ERROR), () =>
-        navigateHome(navigation)
-      );
+      showErrorAlert(error, () => navigateHome(navigation));
```
dalsontws, line 97: For server errors, we can `navigateHome` since it might be connectivity related and user can just retry immediately. However for `sessionError`, we should move the user to the `CampaignLocationScreen` instead? Can refer to how it's done for in `CustomerQuotaScreen.ts`
e.g.
```
 if (error instanceof SessionError) {
        clearCartError();
        expireSession();
        showErrorAlert(error, () => {
          navigation.navigate("CampaignLocationsScreen");
        });
        return;
      }
```

dalsontws (approved):
LGTM 🥳

PR #545, opened 2022-01-25
build: change lint step to use GITHUB_TOKEN

ryanoolala:
changing the lint action to use default GITHUB_TOKEN as it is sufficient for it, no need to provide it with our bot token


dalsontws (approved):
LGTM thanks!!! 🥳

PR #569, opened 2022-03-04
fix: add timeout when fetching data

yao-peijun:
[Notion link](https://www.notion.so/sally-wallet/Handling-FE-Errors-when-Offline-b296a4620ca04f469c3d144346ba9670) 

This PR add timeout when fetching data to fix infinite loading screen when offline.

- [X] I've kept this PR as small as possible (~600 lines) by splitting it into PRs with manageable chunks of code
- [X] I've requested reviews from 1-2 reviewers

### DIFF
--- a/src/services/helpers.ts
+++ b/src/services/helpers.ts
@@ -37,11 +37,20 @@ export async function fetchWithValidator<T, O, I>(
   init?: RequestInit
 ): Promise<T> {
   let response: Response;
+  /*
+   * AbortController() is supported globally for Jest with Node v15 and Jest v27 and above.
+   * Thus, logic is added to check if AbortController is valid for Jest testing purpose only.
+   * TODO: Remove the logic check when node and jest version is compatible.
+   */
   try {
-    const controller = new AbortController();
-    const signal = controller.signal;
+    const controller = typeof AbortController
+      ? {
+          signal: undefined,
+          abort: () => {},
+        }
+      : new AbortController();
     response = await Promise.race<Response>([
-      fetch(requestInfo, { ...init, signal }),
+      fetch(requestInfo, { ...init, signal: controller.signal }),
       timeoutAfter(10).then((timeoutError) => {
         controller.abort();
         throw timeoutError;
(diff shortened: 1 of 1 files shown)


src/services/helpers.ts
```diff
@@ -23,14 +23,27 @@ export class NetworkError extends Error {
   }
 }
 
+const timeoutAfter = (seconds: number): Promise<Error> => {
```
dalsontws, line 26: just curious, was looking [around](https://stackoverflow.com/questions/42147733/doing-a-timeout-error-with-fetch-react-native/54208009) and `AbortController` was talked about as a recommended way to abort `fetch`,  maybe you can share why you decided to use `Promise.race` and a custom timeout instead?

yao-peijun: abortController() aborts the connections and Promise.race returns fetch data or throw specify error after x seconds which ever that is earlier. 

updated the code to abort connection properly after timeout as recommended.

src/services/helpers.ts
```diff
@@ -23,14 +23,40 @@ export class NetworkError extends Error {
...
 ): Promise<T> {
   let response: Response;
+  /*
+   * AbortController() is supported globally for Jest with Node v15 and Jest v27 and above.
+   * Thus, logic is added to check if AbortController is valid for Jest testing purpose only.
+   * TODO: Remove the logic check when node and jest version is compatible.
+   */
   try {
-    response = await fetch(requestInfo, init);
+    const controller = typeof AbortController
```
dalsontws, line 46: why do we need the check here for `typeof AbortController`? can we just `const controller = new AbortController();`?

yao-peijun: Logic added mainly for unit-testing purpose only as AbortController() is not support globally for Node v12.22.1 and Jest v26.6.3 which cause the unit-testing to fail in [92aea07](https://github.com/rationally-app/mobile-application/pull/569/commits/92aea0710ffb30e175604da1ae550efab08751ab). AbortController() is supported globally from Node v15 and Jest v27. [reference](https://github.com/facebook/jest/pull/11182)

dalsontws (approved):
thanks for explaining and taking the time to research on this, LGTM!

PR #606, opened 2022-06-20
feat: add hook to call backend API that checks GovWallet balance

yongjiajie:
[Notion link](https://www.notion.so/sally-wallet/Allow-Sally-to-check-balance-through-GovWallet-v2-customer-account-get-for-MINDEF-SAFRA-campaign-Ha-2fc55b54f9f94053b1a3f71997c2c44f) 

This PR adds... 

- Service that calls API that fetches GovWallet balance
- React Hook `useGovWalletBalance` that manages states based on customer's balance
- Feature Toggle named `checkGovWalletBalance`, which is a `boolean` that determines if we check a customer's balance to determine their eligibility
- Error mesages to support new Error types
- Error handling for new React Hook error states


src/hooks/govwallet/useGovWalletBalance/useGovWalletBalance.test.tsx
```diff
@@ -0,0 +1,158 @@
...
+    );
+
+    expect(result.current.govWalletBalanceState).toStrictEqual("DEFAULT");
+    expect(mockGetGovWalletBalance).toHaveBeenCalledTimes(0);
+  });
+});
+
+describe("govWalletBalanceState states", () => {
+  beforeAll(() => {
+    ids = ["S0000000J", "S0000001I"];
```
Neurrone, line 89: Question: when are there multiple ids?

src/hooks/govwallet/useGovWalletBalance/useGovWalletBalance.tsx
```diff
@@ -0,0 +1,84 @@
...
+          const getBalancePromises = ids.flatMap((id) =>
+            getGovWalletBalance(id, authKey, endpoint)
+          );
+
+          const results = await Promise.all(getBalancePromises);
+
+          // We only retrieve the balance of the first account
+          const areAllBalancesSufficient = results.every(
+            // Check if balance is equals to 10000 cents
+            ({ accountDetails }) => accountDetails[0].balance >= 10000
```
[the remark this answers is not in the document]
yongjiajie: I checked in with Pat, we should be checking for strictly equal to 10000 cents, I will change this!

src/hooks/govwallet/useGovWalletBalance/useGovWalletBalance.test.tsx
yongjiajie: There can be multiple IDs when the campaign allows for checkout for multiple users at the same time. Since this is not the case, we can simplify the tests!

src/hooks/govwallet/useGovWalletBalance/useGovWalletBalance.tsx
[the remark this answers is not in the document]
yongjiajie: Changed this in [774d996](https://github.com/rationally-app/mobile-application/pull/606/commits/774d996e550780ccea7833e9bc70dba9a5e97a1b).

We now check for `balance` to be strictly equal to 10000 cents.

src/hooks/govwallet/useGovWalletBalance/useGovWalletBalance.test.tsx
yongjiajie: Currently, the hook loops through the `ids` and retrieves the balance for each of them, do you think we should simplify it and only retrieve the balance for the first id or keep it as is?

I vote to keep it as is since it is already implemented 🙈

src/hooks/govwallet/useGovWalletBalance/useGovWalletBalance.tsx
```diff
@@ -0,0 +1,84 @@
...
+          const error = e as Error;
+          setGovWalletBalanceError(error);
+        }
+      };
+
+      update();
+    }, [ids, authKey, endpoint]);
+
+  useEffect(() => {
+    if (features?.checkGovWalletBalance && !isEqual(prevIds, ids)) {
```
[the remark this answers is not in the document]
yongjiajie: We compare `prevIds` since we should check the balances of each `id` when they change during rerender. However, I noticed that the `ids` cannot change unless we navigate back out and re-enter this page, thus loading the hook again with new `ids`, hence this is redundant. 

Have removed this in [d85576c](https://github.com/rationally-app/mobile-application/pull/606/commits/d85576c6d2cf4b79f860273fff4cd90d21cf9497)!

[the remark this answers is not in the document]
yongjiajie: As discussed, hooks can only be called from the top-level and should not be called conditionally (as per [this page](https://reactjs.org/docs/hooks-rules.html)), so this is the next best thing 👀

src/hooks/govwallet/useGovWalletBalance/useGovWalletBalance.test.tsx
Neurrone: Ok to leave as is.

src/hooks/govwallet/useGovWalletBalance/useGovWalletBalance.tsx
```diff
@@ -0,0 +1,81 @@
+import { useCallback, useContext, useEffect, useState } from "react";
+import { CampaignConfigContext } from "../../../context/campaignConfig";
+import { getGovWalletBalance } from "../../../services/govwallet/balance";
+
+export type GovWalletBalanceState = "DEFAULT" | "SUFFICIENT" | "INSUFFICIENT";
```
Neurrone, line 5: Because the condition is exactly equal to 10000, do you think changing this to eligible or ineligible would be better? Since 10001 is technically "sufficient"

yongjiajie: Right, that makes sense, renamed `SUFFICIENT` to `ELIGIBLE`, and `INSUFFICIENT` to `INELIGIBLE` in [e4aeff8](https://github.com/rationally-app/mobile-application/pull/606/commits/e4aeff8d70c5f173d63d6cf25e700e3118e56a7e)!

Neurrone: Thanks!

PR #676, opened 2022-10-28
fix: workflow and package versions

wynn987:
[Notion link](https://www.notion.so/sally-wallet/07-Set-up-EAS-build-for-staging-and-prod-8c50e716f11a48edbc4606c8dbe62415) 

This PR does 2 things
1. add a script to insert //npm.pkg.github.com/:_authToken=${NPM_TOKEN} dynamically in ci before any eas commands
- this is only used by eas because it doesnt know our github secrets `NPM_TOKEN`
- having the line locally will cause your npm commands to fail unless you have supplyallybot's PAT. we rotate it often and we don't store it so its troublesome to have it locally. Just add the line dynamically to retrieve it when its available in github secrets for ci purposes
2.  remove the auto increment of package versions because we've been encountering breaking changes even in minor version bumps. Given that we're gonna be talking about whether or not we want to migrate to a web app and also that nothing else can be developed until this story is done, I think we can dont increment for now
- random type check failure https://github.com/rationally-app/mobile-application/actions/runs/3216952883/jobs/5259380826
- random bugged navigation https://www.notion.so/sally-wallet/07-Set-up-EAS-build-for-staging-and-prod-8c50e716f11a48edbc4606c8dbe62415#be7fd2fdc41a412cbbaf4bce1f3c97b8


.github/workflows/test_development_environment.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, macOS-latest, windows-latest]
+        os: [ubuntu-latest, macOS-latest] # temporarily removed windows-latest as it seems to be unable to set token
```
wynn987, line 17: the latest windows os container seems to be unable to get npm token. have checked with windows user dickson to see that the repo is able to run fine on windows outside of the workflow

.github/scripts/inject-token.sh
```diff
@@ -0,0 +1 @@
+echo '//npm.pkg.github.com/:_authToken=${NPM_TOKEN}' | cat - .npmrc > temp && mv temp .npmrc
```
Neurrone, line 1: Is this adding the token to the last line of .npmrc?

.github/workflows/test_development_environment.yml
wynn987: UHH wait ah seems like i was able to get it to pass by running it once without cache and then running it again

Neurrone: Probably fine to remove this, not sure how to get the .sh to work there.

.github/scripts/inject-token.sh
wynn987: to the first line

Neurrone (approved):

PR #733, opened 2023-05-03
chore: try upgrade to expo sdk 46

whoiskai:
[Notion link](https://www.notion.so/sally-wallet/05-Upgrade-to-EXPO-SDK-46-from-SDK-45-096eed02adcc492eb5d9a86d57159fc4?pvs=4) 

This PR adds... 
- update expo sdk from 45 to 46


src/components/CustomerQuota/CustomerCard.tsx
```diff
@@ -76,11 +76,13 @@ export const AddButton: FunctionComponent<AddButton> = ({ text, onPress }) => {
   );
 };
 
-export const CustomerCard: FunctionComponent<{
-  ids: string[];
-  onAddId?: () => void;
-  headerBackgroundColor?: ViewStyle["backgroundColor"];
-}> = ({ ids, onAddId, headerBackgroundColor, children }) => {
+export const CustomerCard: FunctionComponent<
+  PropsWithChildren<{
```
wynn987, line 80: changes across 40 files but its all related to this change. previously `children` can be implicitly defined but now typescript wants us to explicitly define it

package.json
```diff
@@ -1,14 +1,14 @@
 {
   "scripts": {
-    "start": "APP_BUILD_VERSION=1 APP_BINARY_VERSION=dev START_STORYBOOK=false expo start -c --dev-client",
-    "start:mock": "APP_BUILD_VERSION=1 APP_BINARY_VERSION=dev MOCK=true expo start",
+    "start": "APP_BUILD_VERSION=1 APP_BINARY_VERSION=dev START_STORYBOOK=false npx expo start -c --dev-client",
```
dalsontws, line 3: just curious why was `npx` added here ah? since expo is already installed, should work without `npx`?

tsconfig.json
```diff
@@ -9,6 +9,7 @@
     "skipLibCheck": true,
     "resolveJsonModule": true,
     "noImplicitAny": true,
+    "useUnknownInCatchVariables": false,
```
dalsontws, line 12: why do we need this?

whoiskai: cause the typescript >4.4.2 upgrade caused the linting to fail. 

Previously it defaults to `any`, after upgrade it defaults to `unknown`, this caused many will throw error cause we cannot access the properties in the error object unless we type guard it or do a check.

[Example of the errors](https://github.com/rationally-app/mobile-application/actions/runs/4869650347/jobs/8684418391)

This is the fastest way to retain the "old" behaviour. We can remove this option (it defaults to true) when we've updated the types; just felt that it maybe out of scope for this.

package.json
whoiskai: `npx` allows you to use the expo installed in this node modules if exists. Else it'll search on the global space.

If don't use `npx` then it'll always try to search your global space. 

- the old non-npx way (global) has been deprecated: https://docs.expo.dev/archive/expo-cli/
- https://docs.expo.dev/more/expo-cli/ <- modern current version

dalsontws (approved):
LGTM!

PR #765, opened 2023-08-14
chore: pin min sdk version

wynn987:
[Notion link](https://www.notion.so/sally-wallet/03-Sally-Annual-PT-High-finding-Only-support-minimum-Android-version-of-9-and-above-MPT-H001-c48bdcf9a5a047b0900bcd52229f0f06?pvs=4) 

This PR adds min sdk version 27 since the story says > 26 https://docs.expo.dev/versions/latest/sdk/build-properties/

we could do environment variables but i tihnk thats more troublesome because we have to change in 2 places - github and eas. 

the proposed work was to change target and compile version. but after reading through, seems like min version is more correctly what we're trying to do


ShengJie13 (approved):

legendac (approved):
LGTM

PR #777, opened 2024-02-22
chore: updated the QR parser to be able to parse new format [SGW-4296]

YEOWEIHNGWHYELAB:
[Notion link](https://www.notion.so/sally-wallet/GovSupply-Part-2-Mobile-App-Scan-Login-QR-to-go-to-our-web-app-s-domain-url-2f08cb2b51934ba586ed9177cb044ae6) 

This PR adds support for the new QR format. 


src/components/Login/utils.test.tsx
```diff
@@ -2,28 +2,55 @@ import { decodeQr } from "./utils";
...
 
   it("should decode the correct qr code and return endpoint and key", () => {
-    expect.assertions(2);
+    expect.assertions(4);
     const code = `{"key": "1e4457bc-f7d0-4329-a344-f0e3c75d8dd4","endpoint": "https://somewhere.com"}`;
-    const { endpoint, key } = decodeQr(code);
-    expect(endpoint).toBe("https://somewhere.com");
-    expect(key).toBe("1e4457bc-f7d0-4329-a344-f0e3c75d8dd4");
+    const url =
+      "https://somewhere.com?key=1e4457bc-f7d0-4329-a344-f0e3c75d8dd4&endpoint=https://somewhere.com";
```
dalsontws, line 16: let's change the base URL or endpoint here so that it's clearer which one is the endpoint

dalsontws:
```suggestion
      "https://this-is-the-base-url.com?key=1e4457bc-f7d0-4329-a344-f0e3c75d8dd4&endpoint=https://endpoint.com";
```

src/components/Login/utils.ts
```diff
@@ -11,9 +11,45 @@ interface DecodedQrResponse {
...
+    } else {
+      const params = parseURLParams(code);
+      parsedCode = {
+        version: "",
+        endpoint: params.endpoint as string,
+        key: params.key as string,
+      };
+    }
+    if (!parsedCode.endpoint && !parsedCode.key)
+      throw new AuthInvalidError("Invalid format");
```
dalsontws, line 52: do we still need this condition check since there are checks below already?

src/components/Login/utils.test.tsx
```diff
@@ -2,28 +2,55 @@ import { decodeQr } from "./utils";
...
     expect(() => decodeQr(syntaxError)).toThrow("Invalid format");
+
+    const jsonNoEndPoint = `{"key": "1e4457bc-f7d0-4329-a344-f0e3c75d8dd4"}`;
+    expect(() => decodeQr(jsonNoEndPoint)).toThrow("No endpoint");
+
+    const jsonNoKey = `{"endpoint": "example"}`;
+    expect(() => decodeQr(jsonNoKey)).toThrow("No key");
+
+    const missingKeyURL =
+      "http:example.com?keys=1e4457bc-f7d0-4329-a344-f0e3c75d8dd4&endpoint=https://somewhere.com";
```
dalsontws, line 43:
```suggestion
      "http://example.com?keys=1e4457bc-f7d0-4329-a344-f0e3c75d8dd4&endpoint=https://somewhere.com";
```

src/components/Login/utils.test.tsx
```diff
@@ -2,28 +2,55 @@ import { decodeQr } from "./utils";
...
+    expect(() => decodeQr(missingKeyURL)).toThrow("No key");
+
+    const missingEndpointURL =
+      "http:example.com?key=1e4457bc-f7d0-4329-a344-f0e3c75d8dd4&endpointed=https://somewhere.com";
```
dalsontws, line 47:
```suggestion
      "http://example.com?key=1e4457bc-f7d0-4329-a344-f0e3c75d8dd4&endpointed=https://somewhere.com";
```

dalsontws (changes requested):

src/components/Login/utils.ts
```diff
@@ -11,15 +11,45 @@ interface DecodedQrResponse {
...
+const isJSON = (str: string): boolean => {
   try {
-    const parsedCode: QrCode = JSON.parse(code);
-    if (!parsedCode.endpoint) throw new AuthInvalidError("No endpoint");
-    if (!parsedCode.key) throw new AuthInvalidError("No key");
-    return { endpoint: parsedCode.endpoint, key: parsedCode.key };
+    JSON.parse(str);
+    return true;
   } catch (e) {
-    if (e.message.includes("token") || e instanceof SyntaxError)
```
dalsontws, line 21: noticed that the `decodeQr` function on line 38 no longer has a try catch, will this `SyntaxError` go undetected, or will the `isJSON` function cover this scenario?

YEOWEIHNGWHYELAB: Previously when it was only JSON, we could check for syntax error as we know JSON must strictly begin with curly open brace, but now with the addition of URL string type, any syntax error or parsing error will result on `isJSON` being interpreted as a URL.

Instead of using error as a way to tell if the string is a JSON or URL, we could perhaps scan the entire string for `{` and `}` character, if these are present, it means that we can try to parse it as a JSON and throw syntax error upon failure. However, if no `{` or `}` is detected, then we parse it as URL.

dalsontws: okay thanks for explaining, while this works, it seems counter-intuitive (e.g. invalid json shouldn't be detected as a URL) and `if (!parsedCode.endpoint && !parsedCode.key)` covering for `SyntaxError` seems more like a side effect rather than a clear intended outcome.

dalsontws (approved):
approving but leaving the issue open on clarity of error handling! 👍

src/components/Login/utils.ts
dalsontws: good to resolve this or at least have some comments explaining how it works

YEOWEIHNGWHYELAB: oh i see, hmm how about making the parser detect for curly brace character `{` in the QR string so that we know it is a JSON first then we can check if it is a syntax error.

dalsontws: hmm that might not be foolproof as well, instead we can parse the QR and check if it's either a 1. JSON or 2. URL, if it's neither we can treat that is an invalid QR

current behaviour:
1. Check if JSON 
2. If not JSON, parse as URL and find endpoint and key
3. If endpoint and key not found, throw error (JSON syntax errors falls under this category, which could go undetected, we might think its a wrong URL)

Improved behaviour 
1. Check if JSON
2. If not JSON, parse and check if URL
3. If not URL, throw error (not URL and not JSON, SyntaxError covered here)

